PR #3, opened 2016-04-07
Added Policy Evaluation Caching tests.

Bitty9:
Signed-off-by: Navyatha bondugula@ge.com

### DIFF
--- a/acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationCachingIT.java
+++ b/acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationCachingIT.java
@@ -1,0 +1,642 @@
+/*******************************************************************************
+ * Copyright 2016 General Electric Company.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package com.ge.predix.integration.test;
+
+import static com.ge.predix.integration.test.SubjectResourceFixture.BOB_V1;
+import static com.ge.predix.integration.test.SubjectResourceFixture.MARISSA_V1;
+import static com.ge.predix.test.utils.PrivilegeHelper.DEFAULT_RESOURCE_IDENTIFIER;
+import static com.ge.predix.test.utils.PrivilegeHelper.DEFAULT_SUBJECT_ID;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashSet;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.core.env.Environment;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.security.oauth2.client.OAuth2RestTemplate;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.testng.AbstractTestNGSpringContextTests;
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.ge.predix.acs.model.Attribute;
+import com.ge.predix.acs.model.Effect;
+import com.ge.predix.acs.rest.BaseResource;
+import com.ge.predix.acs.rest.BaseSubject;
+import com.ge.predix.acs.rest.PolicyEvaluationRequestV1;
+import com.ge.predix.acs.rest.PolicyEvaluationResult;
+import com.ge.predix.acs.utils.JsonUtils;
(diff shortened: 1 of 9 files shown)


acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationCachingIT.java
```diff
@@ -0,0 +1,519 @@
...
+            responseBody = postForEntity.getBody();
+            Assert.assertEquals(responseBody.getEffect(), Effect.DENY);
+        } finally {
+            if (testPolicyName != null) {
+                this.policyHelper.deletePolicySet(this.acsAdminRestTemplate, this.zoneHelper.getZone1Url(),
+                        testPolicyName, this.acsZone1Headers);
+            }
+        }
+    }
+
```
irinaepshteyn, line 207: What's the difference between testPolicyEvalCacheWhenPolicySetNameChanges() and testPolicyEvalCacheWhenPolicySetChanges().  Both test for changes in the policy set.  Do we want to test just the policy set name?

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationCachingIT.java
```diff
@@ -0,0 +1,519 @@
...
+        } finally {
+            if (testPolicyName != null) {
+                this.policyHelper.deletePolicySet(this.acsAdminRestTemplate, this.zoneHelper.getZone1Url(),
+                        testPolicyName, this.acsZone1Headers);
+                this.privilegeHelper.deleteSubject(this.acsAdminRestTemplate, this.zoneHelper.getZone1Url(),
+                        subject.getSubjectIdentifier(), this.acsZone1Headers);
+            }
+        }
+    }
+
```
irinaepshteyn, line 307: Should we test for resource deletion as well? 
Since API to delete resource is different than API for upsert, we should test cache invalidation for resource deletion.

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationCachingIT.java
irinaepshteyn, line 207: Should we test for deletion of a policy set?

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationCachingIT.java
```diff
@@ -0,0 +1,519 @@
...
```
irinaepshteyn, line 407: Should we also test for a single URI template?  Test that cached decision is invalidated if resource is updated that matches a single URI template...

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationCachingIT.java
```diff
@@ -0,0 +1,519 @@
...
```
irinaepshteyn, line 481: (Minor) Typo. Rename testPolicyEvalCacheWhenResourcesChanges() to testPolicyEvalCacheWhenResourcesChange()

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationCachingIT.java
```diff
@@ -0,0 +1,519 @@
...
```
irinaepshteyn, line 527: Should we test for delete subject and update multiple subjects (postMultipleSubjects in PrivilegeHelper)?
Since API to delete subject is different than API for upsert, we should test cache invalidation for subject deletion. 
Since API to update multiple subjects is different than API to update a single subject, we should test cache invalidation for subject batch update.

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationCachingIT.java
```diff
@@ -0,0 +1,519 @@
...
+                        Effect.DENY } };
+        return data;
+    }
+
+    @DataProvider(name = "endpointProvider")
+    public Object[][] getAcsEndpoint() {
+        Object[][] data = new Object[][] { { this.acsUrl, this.acsZone1Headers }, { this.acsZone1Url, null } };
+        return data;
+    }
+}
```
irinaepshteyn, line 642: Data providers are not used for any of the tests.  These data providers could be used to refactor tests in AccessControlServiceIT

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationCachingIT.java
Bitty9, line 407: We didn't test for single URI template match because testing for two matches covers both for one match and for multiple matches.

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationCachingIT.java
Bitty9, line 642: We should remove these from this class as they are not needed.

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationCachingIT.java
Bitty9, line 207: The only difference is that for testPolicyEvalCacheWhenPolicySetNameChanges() the policy set name is also changed (meaning the initial policy saved in the cache is under a different entry and the "updated" policy is saved in the cache as a new entry). This seems to cover delete for a policy set, unless you mean not have a policy set for a particular acs zone.

irinaepshteyn, line 207: Policy set name is used to derive part of the compound decision key.  testPolicyEvalCacheWhenPolicySetChanges() should keep the same policy set name and change the policy it contains.  testPolicyEvalCacheWhenPolicySetNameChanges() should create a new policy set with the new name.

PR #9, opened 2016-05-06
Disable DEBUG logging for acs to improve performance

sivabalans:
Removed DEBUG flag from logback.xml and application-performance.properties file.

### DIFF
--- a/service/src/main/resources/logback.xml
+++ b/service/src/main/resources/logback.xml
@@ -45,6 +45,8 @@
         level="ERROR" />
     <logger name="springfox.documentation" level="ERROR" />
 
+    <!-- Note: There is severe performance penalty for turning on DEBUG messages during policy evaluation. Limit
+         use for development. -->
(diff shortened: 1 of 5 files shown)


service/src/main/resources/logback.xml
```diff
@@ -45,7 +45,7 @@
         level="ERROR" />
     <logger name="springfox.documentation" level="ERROR" />
 
-    <logger name="com.ge.predix.acs" level="DEBUG" additivity="false">
```
sanjeevchopra, line 48: Maybe we should only disable DEBUG for messages which are printed per policy for now ?  Or log 1 message as INFO per policy evaluation which summarises:
- which policy matched (if there was a match)
- if resource or subject attributes were found (values are confidential)
- what was the result

Especially because we have no way to changing log levels at runtime. It would be good to still have some high level information on evaluation for troubleshooting in production.

PR #18, opened 2016-07-21
Added integration tests for verifying migrated attributes from postgr…

sanjeevchopra:
…es to titan.

Signed-off-by: Bharath bharath.sekar@ge.com

### DIFF
--- a/acs-integration-tests/pom.xml
+++ b/acs-integration-tests/pom.xml
@@ -509,7 +509,7 @@
                         <artifactId>maven-failsafe-plugin</artifactId>
                         <configuration>
                             <forkCount>0</forkCount>
-                            <parallel>false</parallel>
+                            <parallel>methods</parallel>
                             <argLine></argLine>
                             <systemPropertyVariables>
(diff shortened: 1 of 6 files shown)


sanjeevchopra:
pull request build - http://sjc1jenkins08.crd.ge.com:8080/job/Predix-Security/job/acs-build-n-test-pullrequest/599/console

acs-integration-tests/pom.xml
```diff
@@ -509,7 +509,7 @@
                         <artifactId>maven-failsafe-plugin</artifactId>
                         <configuration>
                             <forkCount>0</forkCount>
-                            <parallel>false</parallel>
+                            <parallel>methods</parallel>
```
sanjeevchopra, line 512: should add a comment here in the pom.xml.

This setting does not expect a boolean (see http://maven.apache.org/surefire/maven-failsafe-plugin/examples/fork-options-and-parallel-execution.html).   Currently all test methods are running concurrently which causes failures in ACSMeteringIT  (see https://rally1.rallydev.com/#/59288076735ud/detail/defect/59347373065).  

'methods' implies that only test methods within a test class are executed concurrently.

PR #20, opened 2016-07-23
SubjectRepositoryProxy and ResourceRepositoryProxy to abstract postgres/titan from PrivilegeService

sanjeevchopra:


sanjeevchopra:
@amiri-ge  please review

sanjeevchopra:
http://sjc1jenkins08.crd.ge.com:8080/job/Predix-Security/job/acs-build-n-test-pullrequest/604/ passed

sanjeevchopra:
http://sjc1jenkins08.crd.ge.com:8080/job/Predix-Security/job/acs-build-n-test-pullrequest/605/console passed

service/src/main/java/com/ge/predix/acs/config/GraphBeanDefinitionRegistryPostProcessor.java
```diff
@@ -24,11 +24,9 @@ public void postProcessBeanFactory(final ConfigurableListableBeanFactory beanFac
...
     public void postProcessBeanDefinitionRegistry(final BeanDefinitionRegistry registry) throws BeansException {
         BeanDefinition resourceRepositoryBeanDefinition = new RootBeanDefinition(GraphResourceRepository.class);
-        registry.registerBeanDefinition("resourceRepository", resourceRepositoryBeanDefinition);
         registry.registerBeanDefinition("resourceHierarchicalRepository", resourceRepositoryBeanDefinition);
 
         BeanDefinition subjectRepositoryBeanDefinition = new RootBeanDefinition(GraphSubjectRepository.class);
-        registry.registerBeanDefinition("subjectRepository", subjectRepositoryBeanDefinition);
         registry.registerBeanDefinition("subjectHierarchicalRepository", subjectRepositoryBeanDefinition);
     }
 }
```
irinaepshteyn, line 32: @sanjeevchopra, @bhuvan-rk,
SubjectHierarchicalRepository and resourceHierarchicalRepository are not used anywhere so it seems like we don't have to register its beans.  However, PrivilegeManagementNoRollbackTest and PrivilegeManagementServiceImplTest fail if these registrations are removed.  Need to figure out what the dependency is and remove/change.

service/src/main/java/com/ge/predix/acs/privilege/management/dao/SubjectHierarchicalRepository.java
```diff
@@ -6,9 +6,8 @@
...
 public interface SubjectHierarchicalRepository {
-    SubjectEntity getByZoneAndSubjectIdentifierAndScopes(final ZoneEntity zone, final String subjectIdentifier,
+    SubjectEntity getInheritedAttributesForScopes(final ZoneEntity zone, final String subjectIdentifier,
             final Set<Attribute> scopes);
 
-    SubjectEntity getByZoneAndSubjectIdentifierWithInheritedAttributes(final ZoneEntity zone,
-            final String subjectIdentifier);
-}
+    SubjectEntity getInheritedAttributes(final ZoneEntity zone, final String subjectIdentifier);
+}
```
irinaepshteyn, line 13: Better names would be getSubjectWithInheritedAttributesFroScopes and getSubjectWithInheritedAttributes

service/src/main/java/com/ge/predix/acs/privilege/management/dao/ResourceHierarchicalRepository.java
```diff
@@ -4,7 +4,7 @@
 
 public interface ResourceHierarchicalRepository {
 
-    ResourceEntity getByZoneAndResourceIdentifierWithInheritedAttributes(final ZoneEntity zone,
+    ResourceEntity getInheritedAttributes(final ZoneEntity zone,
             final String resourceIdentifier);
 
 }
```
irinaepshteyn, line 10: Better name would be getResourceWithInheritedAttributes

service/src/main/java/com/ge/predix/acs/config/GraphBeanDefinitionRegistryPostProcessor.java
sanjeevchopra, line 32: @irinaepshteyn  subjectHierarchicalRepository (and resource) are THE hierarchical repositories . They are used by their respective proxy repositories.  (just not wired by name)

service/src/main/java/com/ge/predix/acs/privilege/management/dao/SubjectHierarchicalRepository.java
sanjeevchopra, line 13: its more explicit, though isnt Subject implied from the interface name , SubjectHierarchicalRepository ?   typical calling code would be:

> subjectHierarchicalRepository.getInheritedAttributesForScopes(...);

service/src/main/java/com/ge/predix/acs/privilege/management/dao/ResourceHierarchicalRepository.java
DarioAmiri, line 10: @irinaepshteyn 
@sanjeevchopra 
The intent of "getByZoneAndResourceIdentifierWithInheritedAttributes" was to follow the Spring Data JPA convention. However, since it's not really Spring Data I suppose we can follow any naming conventions.

"getInheritedAttributes" is a bit misleading in that it implies that it is returning something else than an actual instance of a resource object.

"getResourceWithInheritedAttributes" is somewhat redundant in that the method signature is clear about what the return type is.

"getWithResourceAttributes" would probably be the more canonical naming convention but is confusing from an English language point of view.

"getLost" is likely what you're both thinking at this point.

I think being more explicit is the lesser of two evils. I would say let's go with "getResourceWithInheritedAttributes".

irinaepshteyn:
Retest this please

PR #24, opened 2016-08-03
US9933:  Add message property to PolicyEvaluationResult and return a valid error when traversal limit exceeded

bsekar:
US9933: Add message property to PolicyEvaluationResult and return a valid error when traversal limit exceeded
- Added test assertion for the expected message when traversal limit is exceeded
- Added test case with subject attribute exceeding limit. Also, added travelsal limit to error message
- Reset Traversal limit at the end of the test

Signed-off-by: Matias Altman matias.altman@ge.com

### DIFF
--- a/service/src/main/java/com/ge/predix/acs/privilege/management/dao/GraphGenericRepository.java
+++ b/service/src/main/java/com/ge/predix/acs/privilege/management/dao/GraphGenericRepository.java
@@ -35,7 +35,6 @@ import com.ge.predix.acs.rest.Parent;
 import com.ge.predix.acs.utils.JsonUtils;
 import com.ge.predix.acs.zone.management.dao.ZoneEntity;
 import com.google.common.collect.Sets;
-import com.thinkaurelius.titan.core.QueryException;
 import com.thinkaurelius.titan.core.SchemaViolationException;
 
 public abstract class GraphGenericRepository<E extends ZonableEntity> implements JpaRepository<E, Long> {
@@ -402,7 +401,7 @@ public abstract class GraphGenericRepository<E extends ZonableEntity> implements
                         attributes.addAll(deserializedAttributes);
                         // This enforces the limit on the count of attributes returned from the traversal, instead of
                         // number of vertices traversed. To do the latter will require traversing the graph twice.
-                        checkTraversalLimitOrFail(attributes);
+                        checkTraversalLimitOrFail(entity, attributes);
                     }
                 });
 
@@ -415,16 +414,18 @@ public abstract class GraphGenericRepository<E extends ZonableEntity> implements
                             Attribute.class);
                     if (deserializedAttributes != null) {
                         attributes.addAll(deserializedAttributes);
-                        checkTraversalLimitOrFail(attributes);
+                        checkTraversalLimitOrFail(entity, attributes);
                     }
                 });
         entity.setAttributes(attributes);
         entity.setAttributesAsJson(JSON_UTILS.serialize(attributes));
     }
 
-    private void checkTraversalLimitOrFail(final Set<Attribute> attributes) {
+    private void checkTraversalLimitOrFail(final E e, final Set<Attribute> attributes) {
         if (attributes.size() > this.traversalLimit) {
-            throw new QueryException("Graph search failed: traversal limit exceeded.");
+            String exceptionMessage = String.format("The number of attributes on this " + e.getEntityType() + " '"
+                    + e.getEntityId() + "' has exceeded the maximum limit of %d", this.traversalLimit);
+            throw new AttributeLimitExceededException(exceptionMessage);
         }
     }
 
--- a/service/src/main/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationServiceImpl.java
+++ b/service/src/main/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationServiceImpl.java
@@ -46,6 +46,7 @@ import com.ge.predix.acs.policy.evaluation.cache.PolicyEvaluationCacheCircuitBre
 import com.ge.predix.acs.policy.evaluation.cache.PolicyEvaluationRequestCacheKey;
 import com.ge.predix.acs.policy.evaluation.cache.PolicyEvaluationRequestCacheKey.Builder;
 import com.ge.predix.acs.privilege.management.PrivilegeManagementService;
+import com.ge.predix.acs.privilege.management.dao.AttributeLimitExceededException;
 import com.ge.predix.acs.rest.BaseSubject;
 import com.ge.predix.acs.rest.PolicyEvaluationRequestV1;
 import com.ge.predix.acs.rest.PolicyEvaluationResult;
@@ -203,17 +204,27 @@ public class PolicyEvaluationServiceImpl implements PolicyEvaluationService {
             }
             result = new PolicyEvaluationResult(effect, new ArrayList<>(subjectAttributes),
                     new ArrayList<>(resourceAttributes), resolvedResourceUris);
+        } catch (AttributeLimitExceededException ae) {
+            result = handlePolicyEvaluationException(policySet, subjectIdentifier, resourceURI, ae);
         } catch (Throwable e) {
-            StringBuilder builder = new StringBuilder();
-            builder.append("Exception occured while evaluating the policy set. Policy Set ID:")
-                    .append(policySet.getName()).append(" subject:").append(subjectIdentifier)
-                    .append(", Resource URI: ").append(resourceURI);
-            LOGGER.error(builder.toString(), e);
-            result = new PolicyEvaluationResult(Effect.INDETERMINATE);
+            result = handlePolicyEvaluationException(policySet, subjectIdentifier, resourceURI, e);
         }
         return result;
     }
 
+    private PolicyEvaluationResult handlePolicyEvaluationException(final PolicySet policySet,
+            final String subjectIdentifier, final String resourceURI, final Throwable e) {
+        PolicyEvaluationResult result = new PolicyEvaluationResult(Effect.INDETERMINATE);
+        StringBuilder logMessage = new StringBuilder();
+        logMessage.append("Exception occured while evaluating the policy set. Policy Set ID:'")
+                .append(policySet.getName()).append("' subject:'").append(subjectIdentifier)
+                .append("', Resource URI: '").append(resourceURI).append("'");
+        if (e instanceof AttributeLimitExceededException) {
+            result.setMessage(e.getMessage());
+        }
+        LOGGER.error(logMessage.toString(), e);
+        return result;
+    }
     /**
      * @param subjectHandler
      * @param resourceHandler
--- a/service/src/test/java/com/ge/predix/acs/privilege/management/dao/GraphResourceRepositoryTest.java
+++ b/service/src/test/java/com/ge/predix/acs/privilege/management/dao/GraphResourceRepositoryTest.java
@@ -50,7 +50,6 @@ import com.ge.predix.acs.model.Attribute;
 import com.ge.predix.acs.rest.Parent;
 import com.ge.predix.acs.utils.JsonUtils;
 import com.ge.predix.acs.zone.management.dao.ZoneEntity;
-import com.thinkaurelius.titan.core.QueryException;
 import com.thinkaurelius.titan.core.SchemaViolationException;
 import com.thinkaurelius.titan.core.TitanFactory;
 
@@ -397,8 +396,8 @@ public class GraphResourceRepositoryTest {
     }
 
     @Test(
-            expectedExceptions = QueryException.class,
-            expectedExceptionsMessageRegExp = "Graph search failed: traversal limit exceeded.")
+            expectedExceptions = AttributeLimitExceededException.class,
+            expectedExceptionsMessageRegExp = "The number of attributes on this resource .* has exceeded the maximum limit of .*")
     public void testSearchAttributesTraversalLimitException() {
         long traversalLimit = 256L;
         try {
--- a/service/src/test/java/com/ge/predix/controller/test/PolicyEvalWithGraphDbControllerIT.java
+++ b/service/src/test/java/com/ge/predix/controller/test/PolicyEvalWithGraphDbControllerIT.java
@@ -7,6 +7,7 @@ import static com.ge.predix.acs.testutils.XFiles.SPECIAL_AGENTS_GROUP_ATTRIBUTE;
 import static com.ge.predix.acs.testutils.XFiles.TOP_SECRET_CLASSIFICATION;
 import static com.ge.predix.acs.testutils.XFiles.createScopedSubjectHierarchy;
 import static com.ge.predix.acs.testutils.XFiles.createSubjectHierarchy;
+import static com.ge.predix.acs.testutils.XFiles.createTwoLevelResourceHierarchy;
 import static com.ge.predix.acs.testutils.XFiles.createThreeLevelResourceHierarchy;
 import static com.ge.predix.acs.testutils.XFiles.createTwoParentResourceHierarchy;
 import static org.hamcrest.MatcherAssert.assertThat;
@@ -18,6 +19,7 @@ import java.util.Arrays;
 import java.util.List;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.core.env.ConfigurableEnvironment;
 import org.springframework.http.MediaType;
 import org.springframework.test.context.ActiveProfiles;
@@ -39,6 +41,8 @@ import com.ge.predix.acs.model.Attribute;
 import com.ge.predix.acs.model.Effect;
 import com.ge.predix.acs.model.PolicySet;
 import com.ge.predix.acs.privilege.management.PrivilegeManagementService;
+import com.ge.predix.acs.privilege.management.dao.GraphSubjectRepository;
+import com.ge.predix.acs.privilege.management.dao.GraphResourceRepository;
 import com.ge.predix.acs.rest.BaseResource;
 import com.ge.predix.acs.rest.BaseSubject;
 import com.ge.predix.acs.rest.PolicyEvaluationRequestV1;
@@ -64,9 +68,18 @@ public class PolicyEvalWithGraphDbControllerIT extends AbstractTestNGSpringConte
     private static final String POLICY_SET_URL = "v1/policy-set";
     private static final String SUBJECT_URL = "v1/subject";
     private static final String RESOURCE_URL = "v1/resource";
+    private static final long TEST_TRAVERSAL_LIMIT = 2;
 
     @Autowired
     private PrivilegeManagementService privilegeManagementService;
+
+    @Qualifier("resourceHierarchicalRepository")
+    @Autowired
+    private GraphResourceRepository graphResourceRepository;
+
+    @Qualifier("subjectHierarchicalRepository")
+    @Autowired
+    private GraphSubjectRepository graphSubjectRepository;
 
     @Autowired
     private WebApplicationContext wac;
@@ -115,6 +128,7 @@ public class PolicyEvalWithGraphDbControllerIT extends AbstractTestNGSpringConte
             final List<BaseResource> resourceHierarchy, final List<BaseSubject> subjectHierarchy,
             final PolicyEvaluationRequestV1 policyEvalRequest, final Effect expectedEffect) throws Exception {
         // Create policy set.
+
         String uri = POLICY_SET_URL + "/" + URLEncoder.encode(testPolicySet.getName(), "UTF-8");
         MockMvcContext putPolicySetContext = this.testUtils.createWACWithCustomPUTRequestBuilder(this.wac,
                 zone.getSubdomain(), uri);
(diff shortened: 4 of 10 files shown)


service/src/main/java/com/ge/predix/acs/privilege/management/dao/GraphGenericRepository.java
```diff
@@ -423,7 +423,9 @@ private void searchAttributesWithScopes(final E entity, final Vertex vertex, fin
 
     private void checkTraversalLimitOrFail(final Set<Attribute> attributes) {
         if (attributes.size() > this.traversalLimit) {
-            throw new QueryException("Graph search failed: traversal limit exceeded.");
+            String exceptionMessage = String.format("Graph search failed: traversal limit of %s exceeded.",
+                    this.traversalLimit);
```
irinaepshteyn: It would be better to return the exact message that will be propagated upstream: "The number of attributes on resource/subject have exceeded the maximum limit of %s."  This way later on you don't have to manipulate or construct another redundant message as written [here](https://github.com/predix/acs/blob/2c0d68993678945a4e5e62b808734f849e1c0055/service/src/main/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationServiceImpl.java#L207-L228).

service/src/main/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationServiceImpl.java
```diff
@@ -203,6 +204,28 @@ private PolicyEvaluationResult evalPolicySet(final PolicySet policySet, final St
...
+                result.setMessage(message);
+            } else {
+                builder = new StringBuilder();
+                builder.append("Exception occured while evaluating the policy set. Policy Set ID:")
+                        .append(policySet.getName()).append(" subject:").append(subjectIdentifier)
+                        .append(", Resource URI: ").append(resourceURI);
+                LOGGER.error(builder.toString(), qe);
+                result = new PolicyEvaluationResult(Effect.INDETERMINATE);
+            }
+            LOGGER.error(builder.toString(), qe);
```
irinaepshteyn: This error message parsing and construction of new message is redundant.  Just append the error message from the exception, i.e., 
-                         builder.append("Exception occured while evaluating the policy set. Policy Set ID:")
  -                        .append(policySet.getName()).append(" subject:").append(subjectIdentifier)
  -                        .append(", Resource URI: ").append(resourceURI).**append(qe.getMessage()**);

And return this message in the result.

In fact we don't need to explicitly catch QueryException.  Inside [catch (Throwable e)](https://github.com/predix/acs/blob/2c0d68993678945a4e5e62b808734f849e1c0055/service/src/main/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationServiceImpl.java#L229-L238) block below if message is present attach it to the final message and return in the result.

service/src/main/java/com/ge/predix/acs/privilege/management/dao/GraphGenericRepository.java
bsekar: We thought about this approach initially, but then this method does not have access to the subject/resource identifier. So, if we just propagate the message upstream, the actual error message sent to the user will not include the subject/resource id. (Unless we add code to catch this exception upstream and add that information). 
For example:
"The number of attributes on this resource/subject /evidence/implant, mulder have exceeded the maximum limit of 256"
vs
"The number of attributes on this resource/subject have exceeded the maximum limit of 256"

service/src/test/java/com/ge/predix/controller/test/PolicyEvalWithGraphDbControllerIT.java
```diff
@@ -97,9 +113,19 @@ public void setup() throws Exception {
         MockSecurityContext.mockSecurityContext(this.testZone1);
         MockAcsRequestContext.mockAcsRequestContext(this.testZone1);
         this.policySet = this.jsonUtils.deserializeFromFile("complete-sample-policy-set-2.json", PolicySet.class);
+        TRAVERSAL_LIMIT = graphSubjectRepository.getTraversalLimit();
+        graphSubjectRepository.setTraversalLimit(TEST_TRAVERSAL_LIMIT);
+        graphResourceRepository.setTraversalLimit(TEST_TRAVERSAL_LIMIT);
         Assert.assertNotNull(this.policySet, "complete-sample-policy-set-2.json file not found or invalid");
     }
 
+    @AfterClass
```
irinaepshteyn: Instead of setting traversalLimit at the beginning of the whole class and then resetting or cleaning-up after the entire class of tests are done, we can simply do this within the single test method we are writing for evaluating traversal-limit message. 

This can be done within the test method itself. The added advantage of doing so is that now these two [blobs](https://github.com/predix/acs/blob/2c0d68993678945a4e5e62b808734f849e1c0055/service/src/test/java/com/ge/predix/acs/testutils/XFiles.java#L133-L181) become unnecessary. Creating four and five level hierarchy for testing traversal limit is a bit excessive.

service/src/test/java/com/ge/predix/controller/test/PolicyEvalWithGraphDbControllerIT.java
```diff
@@ -178,7 +211,7 @@ public void testPolicyEvaluation(final Zone zone, final PolicySet testPolicySet,
     Object[] scopedAttributeInheritanceData() {
         return new Object[] { this.testZone1, this.policySet, createTwoParentResourceHierarchy(),
                 createScopedSubjectHierarchy(), createPolicyEvalRequest("GET", EVIDENCE_IMPLANT_ID, AGENT_MULDER),
-                Effect.DENY };
+                Effect.DENY, null };
```
irinaepshteyn: We don't need to add an additional parameter that affects all the tests in the class. We can create a separate data-provider object for the traversal-limit tests if needed.

bsekar:
@bhuvan-rk Can you please take a look at this?

bhuvan-rk:
looks good. 👍

service/src/main/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationServiceImpl.java
```diff
@@ -57,6 +57,7 @@
 import com.ge.predix.acs.service.policy.validation.PolicySetValidator;
 import com.ge.predix.acs.zone.management.dao.ZoneEntity;
 import com.ge.predix.acs.zone.resolver.ZoneResolver;
+import com.thinkaurelius.titan.core.QueryException;
```
sanjeevchopra: This violates abstraction - this class should not know/depend on titan classes.  GraphGenericRepository#getEntityWithInheritedAttributes should throw a new exception type (in its package), say  AttributeTraversalLimitException which can be handled in this class.

service/src/main/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationServiceImpl.java
```diff
@@ -203,6 +204,18 @@ private PolicyEvaluationResult evalPolicySet(final PolicySet policySet, final St
             }
             result = new PolicyEvaluationResult(effect, new ArrayList<>(subjectAttributes),
                     new ArrayList<>(resourceAttributes), resolvedResourceUris);
+        } catch (QueryException qe) {
+            result = new PolicyEvaluationResult(Effect.INDETERMINATE);
```
sanjeevchopra: Please move this code block into its own method. This method is very long already (needs to be refacored in a separate story)

service/src/test/java/com/ge/predix/controller/test/PolicyEvalWithGraphDbControllerIT.java
```diff
@@ -64,10 +68,20 @@
     private static final String POLICY_SET_URL = "v1/policy-set";
     private static final String SUBJECT_URL = "v1/subject";
     private static final String RESOURCE_URL = "v1/resource";
+    private static long TRAVERSAL_LIMIT;
```
sanjeevchopra: this is not a constant, just use a method variable in the only test method its used.

service/src/test/java/com/ge/predix/controller/test/PolicyEvalWithGraphDbControllerIT.java
```diff
@@ -154,12 +170,82 @@ public void testPolicyEvaluation(final Zone zone, final PolicySet testPolicySet,
         assertThat(policyEvalResult.getEffect(), equalTo(expectedEffect));
     }
 
+    @Test(dataProvider = "policyEvalExceedingAttributeLimitDataProvider")
```
sanjeevchopra, line 171: this test is only run against one data set. Will be easier to read if there were no testng dataprovider.

service/src/test/java/com/ge/predix/controller/test/PolicyEvalWithGraphDbControllerIT.java
```diff
@@ -154,12 +170,82 @@ public void testPolicyEvaluation(final Zone zone, final PolicySet testPolicySet,
...
+
+        String uri = POLICY_SET_URL + "/" + URLEncoder.encode(testPolicySet.getName(), "UTF-8");
+        MockMvcContext putPolicySetContext = this.testUtils.createWACWithCustomPUTRequestBuilder(this.wac,
+                zone.getSubdomain(), uri);
+        putPolicySetContext.getMockMvc().perform(putPolicySetContext.getBuilder()
+                .contentType(MediaType.APPLICATION_JSON).content(OBJECT_WRITER.writeValueAsString(testPolicySet)))
+                .andExpect(status().isCreated());
+
+        // Create resource hierarchy.
+        if (null != resourceHierarchy) {
```
sanjeevchopra: unnecessary null check.

service/src/test/java/com/ge/predix/controller/test/PolicyEvalWithGraphDbControllerIT.java
```diff
@@ -154,12 +170,82 @@ public void testPolicyEvaluation(final Zone zone, final PolicySet testPolicySet,
...
+            MockMvcContext postResourcesContext = this.testUtils.createWACWithCustomPOSTRequestBuilder(this.wac,
+                    zone.getSubdomain(), RESOURCE_URL);
+            postResourcesContext.getMockMvc()
+                    .perform(postResourcesContext.getBuilder().contentType(MediaType.APPLICATION_JSON)
+                            .content(OBJECT_MAPPER.writeValueAsString(resourceHierarchy)))
+                    .andExpect(status().isNoContent());
+        }
+
+        // Create subject hierarchy.
+        if (null != subjectHierarchy) {
```
sanjeevchopra: unnecessary null check.

service/src/test/java/com/ge/predix/controller/test/PolicyEvalWithGraphDbControllerIT.java
```diff
@@ -154,12 +170,82 @@ public void testPolicyEvaluation(final Zone zone, final PolicySet testPolicySet,
...
+        String expectedMessage;
+        StringBuilder builder = new StringBuilder();
+        builder.append("Exception occured while evaluating the policy set. Policy Set ID:").append(policySet.getName())
+                .append(" subject:").append(policyEvalRequest.getSubjectIdentifier()).append(", Resource URI: ")
+                .append(policyEvalRequest.getResourceIdentifier())
+                .append(", Error is: The number of attributes on this resource/subject has exceeded the maximum limit of "
+                        + TEST_TRAVERSAL_LIMIT);
+        expectedMessage = builder.toString();
+        System.out.println(policyEvalRequest.getResourceIdentifier() + " " + policyEvalRequest.getSubjectIdentifier());
+        if (expectedMessage != null) {
```
sanjeevchopra: unnecessary null check.

service/src/test/java/com/ge/predix/acs/privilege/management/dao/GraphResourceRepositoryTest.java
```diff
@@ -398,7 +398,7 @@ public void testPerformance() throws Exception {
 
     @Test(
             expectedExceptions = QueryException.class,
-            expectedExceptionsMessageRegExp = "Graph search failed: traversal limit exceeded.")
+            expectedExceptionsMessageRegExp = "The number of attributes on this resource/subject has exceeded the maximum limit of .*")
```
sanjeevchopra: Ideally, the message should identify if its the resource or subject whose attributes exceed the limit. I am ok to defer this change, given the low probability of this error. @ilipkind  ?

sanjeevchopra:
Done reviewing - branch needs some changes per comments.

service/src/test/java/com/ge/predix/controller/test/PolicyEvalWithGraphDbControllerIT.java
```diff
@@ -154,12 +170,82 @@ public void testPolicyEvaluation(final Zone zone, final PolicySet testPolicySet,
...
+
+        String expectedMessage;
+        StringBuilder builder = new StringBuilder();
+        builder.append("Exception occured while evaluating the policy set. Policy Set ID:").append(policySet.getName())
+                .append(" subject:").append(policyEvalRequest.getSubjectIdentifier()).append(", Resource URI: ")
+                .append(policyEvalRequest.getResourceIdentifier())
+                .append(", Error is: The number of attributes on this resource/subject has exceeded the maximum limit of "
+                        + TEST_TRAVERSAL_LIMIT);
+        expectedMessage = builder.toString();
+        System.out.println(policyEvalRequest.getResourceIdentifier() + " " + policyEvalRequest.getSubjectIdentifier());
```
sanjeevchopra: pls remove

service/src/main/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationServiceImpl.java
```diff
@@ -203,17 +204,27 @@ private PolicyEvaluationResult evalPolicySet(final PolicySet policySet, final St
...
+        } catch (AttributeTraversalException ae) {
+            result = setIndeterminateMessage(policySet, subjectIdentifier, resourceURI, ae.getMessage());
         } catch (Throwable e) {
-            StringBuilder builder = new StringBuilder();
-            builder.append("Exception occured while evaluating the policy set. Policy Set ID:")
-                    .append(policySet.getName()).append(" subject:").append(subjectIdentifier)
-                    .append(", Resource URI: ").append(resourceURI);
-            LOGGER.error(builder.toString(), e);
-            result = new PolicyEvaluationResult(Effect.INDETERMINATE);
+            result = setIndeterminateMessage(policySet, subjectIdentifier, resourceURI, null);
```
sanjeevchopra: after this change, any unexpected exceptions will not be logged.  lets discuss.

service/src/main/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationServiceImpl.java
```diff
@@ -203,17 +204,27 @@ private PolicyEvaluationResult evalPolicySet(final PolicySet policySet, final St
...
         }
         return result;
     }
 
+    private PolicyEvaluationResult setIndeterminateMessage(final PolicySet policySet, final String subjectIdentifier,
```
sanjeevchopra: better method name might be handlePolicyEvaluationException

service/src/test/java/com/ge/predix/controller/test/PolicyEvalWithGraphDbControllerIT.java
```diff
@@ -152,6 +217,9 @@ public void testPolicyEvaluation(final Zone zone, final PolicySet testPolicySet,
         PolicyEvaluationResult policyEvalResult = OBJECT_MAPPER
                 .readValue(mvcResult.getResponse().getContentAsByteArray(), PolicyEvaluationResult.class);
         assertThat(policyEvalResult.getEffect(), equalTo(expectedEffect));
+        assertThat(policyEvalResult.getMessage(), equalTo(expectedMessage));
+        graphResourceRepository.setTraversalLimit(traversalLimit);
```
sanjeevchopra: to be safe, add the reverts of limits in a finally clause.

service/src/test/java/com/ge/predix/controller/test/PolicyEvalWithGraphDbControllerIT.java
```diff
@@ -202,6 +276,30 @@ public void testPolicyEvaluation(final Zone zone, final PolicySet testPolicySet,
                 Effect.PERMIT };
     }
 
+    /**
+     * Test that evaluation is successful when the resource and/or subject attributes exceed the length. The policy set will return
+     * indeterminate because the traversal limit is exceeded.
+     */
+    Object[] evaluationWithResourceAttributesExceedingTraversalLimitData() {
+        String errorMessage = "The number of attributes on this resource "
```
sanjeevchopra: "The number of attributes on resource '<resource id>' exceeds the maximum of <limit>."

service/src/main/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationServiceImpl.java
```diff
@@ -203,17 +204,27 @@ private PolicyEvaluationResult evalPolicySet(final PolicySet policySet, final St
...
+            final String resourceURI, final String message) {
+        PolicyEvaluationResult result = new PolicyEvaluationResult(Effect.INDETERMINATE);
+        StringBuilder builder = new StringBuilder();
```
sanjeevchopra: rename 'builder' to 'logMessage' ?

service/src/test/java/com/ge/predix/controller/test/PolicyEvalWithGraphDbControllerIT.java
```diff
@@ -202,6 +276,30 @@ public void testPolicyEvaluation(final Zone zone, final PolicySet testPolicySet,
...
+                createPolicyEvalRequest("GET", EVIDENCE_SCULLYS_TESTIMONY_ID, AGENT_MULDER),
+                Effect.INDETERMINATE, errorMessage};
+    }
+
+    /**
+     * Test that subjects and resources inherit attributes from their parents. The policy set will permit the request
+     * if the subject and resource successfully inherit the required attributes from their respective parents.
+     */
+    Object[] evaluationWithSubjectAttributesExceedingTraversalLimitData() {
+        String errorMessage = "The number of attributes on this subject "
```
sanjeevchopra: "The number of attributes on subject '<subject id>' exceeds the maximum of <limit>."

service/src/main/java/com/ge/predix/acs/privilege/management/dao/AttributeTraversalException.java
```diff
@@ -0,0 +1,14 @@
+package com.ge.predix.acs.privilege.management.dao;
+
+public class AttributeTraversalException extends RuntimeException {
```
DarioAmiri: Semantically speaking, we are traversing a graph - not attributes. Maybe GraphTraversalException would make more sense? Graphs are well-known data structures so users might be able to understand graph traversal, a well-known problem in CS, than attribute traversal (something we more or less made up).

sanjeevchopra: How about AttributeResolutionException ?  Also, note that we are not exposing this to our users.

service/src/main/java/com/ge/predix/acs/privilege/management/dao/GraphGenericRepository.java
```diff
@@ -415,16 +414,18 @@ private void searchAttributesWithScopes(final E entity, final Vertex vertex, fin
...
+                        checkTraversalLimitOrFail(entity, attributes);
                     }
                 });
         entity.setAttributes(attributes);
         entity.setAttributesAsJson(JSON_UTILS.serialize(attributes));
     }
 
-    private void checkTraversalLimitOrFail(final Set<Attribute> attributes) {
+    private void checkTraversalLimitOrFail(final E e, final Set<Attribute> attributes) {
         if (attributes.size() > this.traversalLimit) {
```
amiri-ge: I don't fully comprehend the code at this point but at first glance this seems to be missing the point. The whole point of the traversal limit is to limit the traversal. Why would we ever allow the attribute size to exceed this?

service/src/main/java/com/ge/predix/acs/privilege/management/dao/GraphGenericRepository.java
```diff
@@ -402,7 +401,7 @@ private void searchAttributesWithScopes(final E entity, final Vertex vertex, fin
                         attributes.addAll(deserializedAttributes);
                         // This enforces the limit on the count of attributes returned from the traversal, instead of
                         // number of vertices traversed. To do the latter will require traversing the graph twice.
-                        checkTraversalLimitOrFail(attributes);
+                        checkTraversalLimitOrFail(entity, attributes);
```
DarioAmiri, line 404: Rather than going about it this way, consider using the `sideEffect()` intermediate in place of subject. `sideEffect()` allows you to inspect the current state of the graph traverser (similar to peek() in Java 8 streams). For example:

`.sideEffect( t -> { if ( t.loops() > limit ) doSomething(); })`

This would limit based on the number of loops the traverser has gone through. There might be ways to limit based on results as well. Not sure, but we should explore other options.

sanjeevchopra, line 404: I will create a new story for this, with these notes.

DarioAmiri, line 404: Disregard my earlier comment. I think after playing with the code I understand the problem a bit better. However, there are still things to consider. First of all, realize that because the code is turned to a stream, no actual work is done by the code until you call `forEach()` which will begin to drive the actual traversal. Once you realize that, you will also realize that the `limit()` is not really entirely needed. You can just keep track of the number of times your `forEach()` lambda is called and simply throw an exception when the limit is reached.

service/src/main/java/com/ge/predix/acs/privilege/management/dao/AttributeTraversalException.java
DarioAmiri: After reading the code a bit more, it seems to me that the exception occurs when a traversal limit is reached. So why not call it something that reflects why it actually occurred. Such as "TraversalLimitException" or similar? What does attribute resolution have to do with it? If I was reading this code and saw that the exception incorporated "limit" or "LimitExceeded" I would have a better understanding of what the exception is for. In other words, name the exception after an event not an operation.

service/src/main/java/com/ge/predix/acs/privilege/management/dao/GraphGenericRepository.java
DarioAmiri, line 404: To add to the previous comment, see [here](https://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html) for a discussion of "intermediate" vs "terminal" operations in Java 8 streams. The bottom line is that streams are lazy. Nothing happens until you call a terminal operations such as `forEach()`, `collect()`, `reduce()`, etc.

Also check out the code for `limit()`. You will see that it is just a `filter()` that applies a range to the result size.

sanjeevchopra, line 404: to be clear, is this what you are proposing ?
- remove limit() from graph traversal
- replace line 404 with:

>  int traversalCount;
>  if (++traversalCount > this .traversalLimit)  {
>     throw new AttributeTraversalLimitExceededException()...
>  }

service/src/main/java/com/ge/predix/acs/privilege/management/dao/AttributeTraversalException.java
sanjeevchopra: agreed, to be explicit lets call it  AttributeLimitExceededException

DarioAmiri: agreed AttributeLimitExceededException is much easier to understand.

service/src/main/java/com/ge/predix/acs/privilege/management/dao/GraphGenericRepository.java
DarioAmiri, line 404: Basically, yes. I think limit is only a good option if you're not interested in raising an event if the limit is actually hit.

service/src/main/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationServiceImpl.java
```diff
@@ -203,17 +204,27 @@ private PolicyEvaluationResult evalPolicySet(final PolicySet policySet, final St
...
+            result = handlePolicyEvaluationException(policySet, subjectIdentifier, resourceURI, e);
         }
         return result;
     }
 
+    private PolicyEvaluationResult handlePolicyEvaluationException(final PolicySet policySet,
+            final String subjectIdentifier, final String resourceURI, final Throwable e) {
+        PolicyEvaluationResult result = new PolicyEvaluationResult(Effect.INDETERMINATE);
+        StringBuilder logMessage = new StringBuilder();
+        logMessage.append("Exception occured while evaluating the policy set. Policy Set ID:")
```
DarioAmiri: There are some missing spaces after semicolons in this message. Not sure if that is intentional. Also consider single quoting the values you are injecting into the string. Makes it easier to read and select the text.

service/src/test/java/com/ge/predix/controller/test/PolicyEvalWithGraphDbControllerIT.java
DarioAmiri, line 171: @sanjeevchopra This might be future-proofing. Do we anticipate that there might be more test scenarios/data sets to run in the future?

service/src/main/java/com/ge/predix/acs/privilege/management/dao/GraphGenericRepository.java
sanjeevchopra, line 404: ok, I created a separate story for this - https://rally1.rallydev.com/#/59288076735ud/detail/userstory/60661742041?fdp=true

sanjeevchopra:
checks passed - http://sjc1jenkins08.crd.ge.com:8080/job/Predix-Security/job/acs-build-n-test-pullrequest/618/

PR #26, opened 2016-08-11
Pull Request for Implementing attribute migration in ACS 

bsekar:
Complete list of changes for migrating attributes from Postgres to Titan, includes vertex-management and the actual migration data flow.


bsekar:
@amiri-ge PR for attribute migration to Titan. FYI.

service/src/main/java/com/ge/predix/acs/privilege/management/dao/TitanMigrationManager.java
```diff
@@ -0,0 +1,66 @@
...
+    public TitanMigrationManager() {
+
+    }
+
+    @PostConstruct
+    public void doMigration() {
+        // TODO: Return value should reflect status of migration?
+        // Success, failure, interrupted etc ?
+        try {
+            if (this.resourceProxy.getVersion() == 0) {
```
sanjeevchopra: * this block is checking for version on the resource graph, however migrating both resource and subject. 
* lets separate the two migrations here. Version on subject graph also needs to be updated to 1

service/src/main/java/com/ge/predix/acs/privilege/management/dao/TitanMigrationManager.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    public TitanMigrationManager() {
+
+    }
+
+    @PostConstruct
+    public void doMigration() {
+        // TODO: Return value should reflect status of migration?
+        // Success, failure, interrupted etc ?
```
sanjeevchopra: Please remove. Any TODOs should be captured in stories.  In this case, I dont see a need for a return code.

service/src/main/java/com/ge/predix/acs/privilege/management/dao/TitanMigrationManager.java
```diff
@@ -0,0 +1,66 @@
...
+
+                List<SubjectEntity> listOfSubjects = subjectProxy.getNonGraphRepository().findAll();
+                LOGGER.info("Found %d subjects to migrate", listOfSubjects.size());
+                subjectProxy.getGraphRepository().save(listOfSubjects);
+                LOGGER.info("Subject migration to Titan completed.");
+                this.resourceProxy.setVersion(1);
+                LOGGER.info("Exiting from Titan attribute migration process.");
+            }
+            return;
+        } catch (Exception e) {
```
sanjeevchopra: catch `Throwable` instead of Exception here, to handle all "throwables".

service/src/main/java/com/ge/predix/acs/privilege/management/dao/ResourceRepositoryProxy.java
```diff
@@ -26,10 +26,18 @@
     @Autowired(required = false)
     private GraphResourceRepository graphRepository;
 
+    public GraphResourceRepository getGraphRepository() {
```
sanjeevchopra: Please make these methods package private - should not be needed by services or any other using "layers".  

On second thought, it may be better to not have these methods at all - just for the benefit of one time migration code.  Can we wire both the repositories directly in the migration manager ? This keeps the proxy interface cleaner. Also, at some future version we might remove the non-graph repository.

service/src/main/java/com/ge/predix/acs/privilege/management/dao/TitanMigrationManager.java
```diff
@@ -0,0 +1,66 @@
...
+
+    }
+
+    @PostConstruct
+    public void doMigration() {
+        // TODO: Return value should reflect status of migration?
+        // Success, failure, interrupted etc ?
+        try {
+            if (this.resourceProxy.getVersion() == 0) {
+                // This is basically bringing all of Postgres data into memory. Should we
```
sanjeevchopra: the non graph repository supports paging.   Please use:

> Page<T> findAll(Pageable pageable);

sanjeevchopra:
done reviewing - branch needs some changes

service/src/main/java/com/ge/predix/acs/privilege/management/dao/ResourceRepositoryProxy.java
bsekar: Valid point about changing this method from being public. So we've made it package private. 

Also, we wanted to keep the proxy class as a single entry-point for all repo related queries. So we avoided auto-wiring 4 individual repos in this class and use the proxy.  We can discuss this

service/src/main/java/com/ge/predix/acs/privilege/management/dao/TitanMigrationManager.java
```diff
@@ -0,0 +1,96 @@
...
+
+                int totalResources = 0;
+                int totalSubjects = 0;
+                Pageable pageRequest = new PageRequest(0, PAGE_SIZE);
+                do {
+                    Page<ResourceEntity> pageOfResources = resourceRepository.findAll(pageRequest);
+                    List<ResourceEntity> listOfResources = pageOfResources.getContent();
+                    totalResources += pageOfResources.getNumberOfElements();
+                    listOfResources.forEach(item -> item.setId((long) 0));
+                    resourceHierarchicalRepository.save(listOfResources);
```
sanjeevchopra: add another log message here, in case there is a exception before all pages are saved.  And keep the log at line 73 which logs the total number. 
same for subject block below.

service/src/test/java/com/ge/predix/acs/privilege/management/dao/TitanMigrationManagerTest.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    private List<SubjectEntity> saveSubjectsToGraph(final Iterable<SubjectEntity> entities) {
+        List<SubjectEntity> savedEntities = new ArrayList<>();
+        entities.forEach(item -> savedEntities.add(item));
+        toListForSubject.addAll(savedEntities);
+        return savedEntities;
+    }
+
+   // @Test
```
sanjeevchopra: is this still work in progress ?

bsekar:
Yes, this is still in progress

service/src/test/java/com/ge/predix/acs/privilege/management/dao/TitanMigrationManagerTest.java
bsekar: yes, this is still in progress

bsekar:
@sanjeevchopra @irinaepshteyn changes for Attribute migration in ACS : Please review.

service/src/test/java/com/ge/predix/acs/privilege/management/dao/TitanMigrationManagerTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+        // Verify that the non-graph repos are populated
+        assertThat(this.resourceRepository.findAll().size(), equalTo(2));
+        assertThat(this.resourceHierarchicalRepository.findAll().size(), equalTo(0));
+
+        assertThat(this.subjectRepository.findAll().size(), equalTo(2));
+        assertThat(this.subjectHierarchicalRepository.findAll().size(), equalTo(0));
+
+        // Mock graph-save function with a locally defined saveResourcesToGraph()
+        Mockito.when(this.resourceHierarchicalRepository.save(fromListForResource))
```
sanjeevchopra: > Mockito.when(this.resourceHierarchicalRepository.save(anyCollectionOf(ResourceEntity.class)))

sanjeevchopra:
tests passed - http://sjc1jenkins08.crd.ge.com:8080/job/Predix-Security/job/acs-build-n-test-pullrequest/623/

PR #53, opened 2016-10-08
US45006: Make subject attributes consistent

matiasa123:
- Modified PolicyEvaluationRequestV1Test.java to reflect Set is unordered
- Modified PolicyEvaluationRequestV1 so that subjectAttributes and resourceAttributes are sets instead of list
- Modified PolicyEvaluationRequestV1 so that functions reflect Subject/ResourceAttributes are sets instead of list
- Modified Unit tests that failed to compile due to the changes to PolicyEval
- Modified Integration tests to reflect new changes

### DIFF
--- a/acs-integration-tests/src/test/java/com/ge/predix/test/utils/PolicyHelper.java
+++ b/acs-integration-tests/src/test/java/com/ge/predix/test/utils/PolicyHelper.java
@@ -20,9 +20,9 @@ import static com.ge.predix.test.utils.ACSTestUtil.ACS_VERSION;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.Collections;
 import java.util.Random;
+import java.util.Set;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpEntity;
@@ -102,13 +102,13 @@ public class PolicyHelper {
 
     public PolicyEvaluationRequestV1 createRandomEvalRequest() {
         Random r = new Random(System.currentTimeMillis());
-        List<Attribute> subjectAttributes = new ArrayList<Attribute>();
+        Set<Attribute> subjectAttributes = Collections.emptySet();
         return this.createEvalRequest(ACTIONS[r.nextInt(4)], String.valueOf(r.nextLong()),
                 "/alarms/sites/" + String.valueOf(r.nextLong()), subjectAttributes);
     }
 
     public PolicyEvaluationRequestV1 createEvalRequest(final String action, final String subjectIdentifier,
-            final String resourceIdentifier, final List<Attribute> subjectAttributes) {
+            final String resourceIdentifier, final Set<Attribute> subjectAttributes) {
         PolicyEvaluationRequestV1 policyEvaluationRequest = new PolicyEvaluationRequestV1();
         policyEvaluationRequest.setAction(action);
         policyEvaluationRequest.setSubjectIdentifier(subjectIdentifier);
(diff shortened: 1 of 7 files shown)


acs-integration-tests/src/test/java/com/ge/predix/test/utils/PolicyHelper.java
```diff
@@ -20,9 +20,7 @@
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Random;
+import java.util.*;
```
DarioAmiri: Wildcard imports are considered a bad practice please be explicit.

DarioAmiri (changes requested):
Clean up the import I pointed out. Everything else looks good.

matiasa123:
http://predix1.jenkins.build.ge.com/job/Predix-Security/job/acs-build-n-test-pullrequest/670/console
Recording test results
Recording fingerprints
Notifying upstream projects of job completion
Finished: SUCCESS

matiasa123:
Fixed thanks for catching that

model/src/main/java/com/ge/predix/acs/rest/PolicyEvaluationRequestV1.java
```diff
@@ -119,9 +119,7 @@ public boolean equals(final Object obj) {
                 return false;
             }
             if ((null != this.subjectAttributes) && (this.subjectAttributes.size() == other.subjectAttributes.size())) {
-                for (int i = 0; i < this.subjectAttributes.size(); i++) {
-                    equalsBuilder.append(this.subjectAttributes.get(i), other.subjectAttributes.get(i));
-                }
+                equalsBuilder.append(this.subjectAttributes, other.subjectAttributes);
```
DarioAmiri, line 127: I don't believe this is the equivalent of comparing all items for equality. HashSet.equals() does not call equals() for every item in the set. See See https://docs.oracle.com/javase/7/docs/api/java/util/Set.html

DarioAmiri (changes requested):

model/src/main/java/com/ge/predix/acs/rest/PolicyEvaluationRequestV1.java
matiasa123: I can see where your concerns come from.  At first I was hesitant to use the built in equals operation with set.  My thinking is it would be wise to use the built in equals because we can't control what kind of set our users are loading into PolicyEvalRequest and there may be equals criteria outside of matching type, size, and elements.  

Also it is important to point out that depending on the set our user uses we may not have any guarantee's with how the set is ordered (http://www.programcreek.com/2013/03/hashset-vs-treeset-vs-linkedhashset/).  This is my primary concern with implementing element by element comparison in our function especially considering we are seeing alot of HashSets.   

According to HashSets documents equals is inherited from AbstractSet
https://docs.oracle.com/javase/7/docs/api/java/util/HashSet.html

AbstractSet equals checks type, set size, and then each individual element.
Documentation says
This implementation first checks if the specified object is this set; if so it returns true. Then, it checks if the specified object is a set whose size is identical to the size of this set; if not, it returns false. If so, it returns containsAll((Collection) o).
https://docs.oracle.com/javase/7/docs/api/java/util/AbstractSet.html#equals(java.lang.Object)

matiasa123: Offline we discussed this and  @DarioAmiri agreed but requested that comments be added for anyone who runs into this line of code later.

DarioAmiri (approved):

PR #69, opened 2016-12-15
This checkin removes dependency of model and commons on hystrix libra…

anurag-gujral:
…ries

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,9 +45,9 @@
         <maven-compiler-plugin.version>2.3.2</maven-compiler-plugin.version>
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
+        <hystrix.version>1.5.2</hystrix.version>
         <maven-bundle-plugin.version>2.3.7</maven-bundle-plugin.version>
         <archaius.version>0.7.1</archaius.version>
-        <hystrix.version>1.5.2</hystrix.version>
         <!-- Spring framework upgrade to 4.3.1 requires this version of Jackson -->
         <jackson.version>2.5.0</jackson.version>
         <jedis.version>2.8.1</jedis.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -132,32 +132,11 @@
...
-            <exclusions>
-                <exclusion>
-                    <groupId>com.netflix.archaius</groupId>
-                    <artifactId>archaius-core</artifactId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-        <dependency>
             <groupId>com.netflix.archaius</groupId>
             <artifactId>archaius-core</artifactId>
```
anubhavi25: we should remove this Hystrix-related dependency if it's not used in non-service modules (it's defined in service/pom.xml so the service code can still use it)

pom.xml
```diff
@@ -132,32 +132,11 @@
...
-            <version>${hystrix.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>com.netflix.hystrix</groupId>
-            <artifactId>hystrix-metrics-event-stream</artifactId>
-            <version>${hystrix.version}</version>
-        </dependency>
-        <dependency>
             <groupId>org.springframework.cloud</groupId>
             <artifactId>spring-cloud-commons</artifactId>
```
anubhavi25: we should remove this Hystrix-related dependency if it's not used in non-service modules (it's defined in service/pom.xml so the service code can still use it). also remove the "spring-cloud-netflix-core" dependency below since it also doesn't seem to be needed.

anubhavi25 (changes requested):

pom.xml
anurag-gujral: This change has been addressed, thanks for noticing this.

pom.xml
anurag-gujral: This change has been addressed, thanks for noticing this.

PR #73, opened 2016-12-21
US63621: Use the Set interface instead of LinkedHashSet where necessary

anubhavi25:


anubhavi25:
[Build Status](https://predix1.jenkins.build.ge.com/job/Predix-Security/job/acs-build-n-test-pullrequest/725/)

irinaepshteyn:
@anubhavi25 , @anurag-gujral , @FrankGasparovic , @sanjeevchopra ,

This refactoring is not justified.  List of policy set IDs passed in evaluation request specifies the order in which the service will evaluate policies.  Though it's mentioned in ACS documentation, I am not sure we could guarantee that ACS model consumers would follow it carefully and remember to create an ordered set on their own.  Explicitly enforcing LinkedHashSet instead of Set in policy evaluation request was a way to enforce the correct usage and guarantee to preserve the original order of policy sets created by ACS model consumers.

anubhavi25:
My 2 cents: while I agree that having `LinkedHashSet` passed in everywhere instead of the `Set` interface makes it clear that insertion-order is what's preferred, it has some disadvantages:

  - It still doesn't enforce a compile-time guarantee. For example, even if a function signature is `public void setPolicySetsEvaluationOrder(final LinkedHashSet<String> policySetIds);`, clients can still shoot themselves in the foot with something like `setPolicySetsEvaluationOrder(new LinkedHashSet<String>(<some_unordered_collection>))` where `<some_unordered_collection>` is passed in from some part of their business logic. At this point, they would have to read the ACS documentation anyways and find that order is significant.
  - It doesn’t allow us to easily change the underlying type of set to, say, a `TreeSet` if we impose a custom ordering based on a `Comparator`.

While I'm leaning towards using the `Set` interface to mitigate the above disadvantages, I don’t have a strong preference either way as long as we reach a consensus.

sanjeevchopra:
Lets discuss this in person.

sanjeevchopra:
discussed with team - and decided to leave this as is.

PR #79, opened 2017-01-19
US70421: Configured the main app and dependent services (UAA, ZAC, et…

anubhavi25:
…c.) to run locally on non-standard ports


service/start-acs-public.sh
```diff
@@ -1,4 +1,5 @@
-#!/bin/bash
```
FrankGasparovic, line 1: Need to update all start-acs scripts with the new set-env-local script

PR #90, opened 2017-02-21
US65644: Map ACS-specific audit events to Predix-specific ones

FrankGasparovic:


service/src/main/java/com/ge/predix/acs/audit/PredixEventMapper.java
```diff
@@ -0,0 +1,44 @@
...
+import com.ge.predix.audit.sdk.message.AuditEnums.CategoryType;
+import com.ge.predix.audit.sdk.message.AuditEnums.Classifier;
+import com.ge.predix.audit.sdk.message.AuditEnums.EventType;
+import com.ge.predix.audit.sdk.message.AuditEnums.PublisherType;
+import com.ge.predix.audit.sdk.message.AuditEventV2;
+import com.ge.predix.audit.sdk.message.AuditEventV2.AuditEventV2Builder;
+
+public class PredixEventMapper {
+
+    public AuditEventV2 map(final com.ge.predix.audit.AuditEvent auditEvent) {
```
anubhavi25: No need to specify the full package name since now there's no name clash between `AuditEvent` and `AuditEventV2`

service/src/main/java/com/ge/predix/acs/audit/PredixEventMapper.java
```diff
@@ -0,0 +1,44 @@
...
+        Map<String, String> auditPayload = new HashMap<String, String>();
```
anubhavi25: You can just use the diamond operator `<>` since you already gave the type specification via the `Map` declaration: `Map<String, String> auditPayload = new HashMap<>();`

service/src/main/java/com/ge/predix/acs/audit/PredixEventMapper.java
```diff
@@ -0,0 +1,44 @@
...
+        } else {
+            auditEventBuilder = auditEventBuilder.eventType(EventType.FAILURE_API_REQUEST)
+                    .classifier(Classifier.FAILURE);
+        }
+
+        return auditEventBuilder.build();
+
+    }
+
+    private boolean isSuccessful(final com.ge.predix.audit.AuditEvent auditEvent) {
```
anubhavi25: See comment re. not needing the full package name

service/src/main/java/com/ge/predix/acs/audit/PredixEventMapper.java
```diff
@@ -0,0 +1,44 @@
...
+    public AuditEventV2 map(final com.ge.predix.audit.AuditEvent auditEvent) {
+        Map<String, String> auditPayload = new HashMap<String, String>();
+        auditPayload.put("httpMethod", auditEvent.getMethod());
+        auditPayload.put("requestBody", auditEvent.getRequestBody());
+        auditPayload.put("responseBody", auditEvent.getResponseBody());
+
+        AuditEventV2Builder auditEventBuilder = AuditEventV2.builder().timestamp(auditEvent.getTime().toEpochMilli())
+                .correlationId(auditEvent.getCorrelationId()).tenantUuid(auditEvent.getZoneId())
+                .publisherType(PublisherType.APP_SERVICE).categoryType(CategoryType.API_CALLS)
+                .payload(auditPayload.toString());
```
anubhavi25: `auditPayload.toString()` will invoke `AbstractMap#toString()` which looks like JSON but isn't. Echoing @anurag-gujral, shouldn't we be serializing this as JSON instead?

service/src/main/java/com/ge/predix/acs/audit/PredixEventProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+import com.ge.predix.audit.sdk.message.AuditEventV2;
+
+public class PredixEventProcessor implements AuditEventProcessor {
+
+    private static final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(PredixEventProcessor.class);
+
+    @Autowired
+    private AuditClient auditClient;
+
+    @Autowired
```
anubhavi25, line 19: This coincidentally works since `PredixEventProcessor` and `PredixEventMapper` are part of the same package, but it might break if trying to wire an instance of `PredixEventMapper` elsewhere. Perhaps add an `@Component` annotation to the `PredixEventMapper` class above?

service/src/main/java/com/ge/predix/acs/audit/PredixEventProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+    private PredixEventMapper eventMapper;
+
+    @Override
+    public boolean process(final com.ge.predix.audit.AuditEvent auditEvent) {
```
anubhavi25: See comment re. not needing the full package name

service/src/test/java/com/ge/predix/acs/audit/PredixEventMapperTest.java
```diff
@@ -0,0 +1,66 @@
...
+        cachedResponse.getWriter().write(responseBody);
+        cachedRequest.getReader().readLine();
+        
+        AuditEvent event = new AuditEvent(cachedRequest, cachedResponse, zoneId, correlationId);
+        AuditEventV2 predixEvent = mapper.map(event);
+        
+        Assert.assertTrue(predixEvent.getPayload().contains(requestBody));
+        Assert.assertTrue(predixEvent.getPayload().contains(methodType));
+        Assert.assertTrue(predixEvent.getPayload().contains(responseBody));
+        Assert.assertEquals(predixEvent.getCategoryType(), categoryType);
```
anubhavi25: Add: `Assert.assertEquals(predixEvent.getPublisherType(), publisherType);`

service/src/test/java/com/ge/predix/acs/audit/PredixEventMapperTest.java
```diff
@@ -0,0 +1,66 @@
...
+    // CHECKSTYLE:ON: MagicNumber
+
+    @DataProvider
+    public Object[][] predixMapperDataProvider() {
+        return new Object[][] {
+                new Object[] { "request body", "POST", 401, "response body", "1234", "5678", PublisherType.APP_SERVICE,
+                        CategoryType.API_CALLS, EventType.FAILURE_API_REQUEST, Classifier.FAILURE },
+                { "request body", "PUT", 200, "response body", "9101112", "13141516", PublisherType.APP_SERVICE,
+                        CategoryType.API_CALLS, EventType.SUCCESS_API_REQUEST, Classifier.SUCCESS } };
+    }
```
anubhavi25, line 64: We should probably add at least one test for auditing input validation failures: `{ ... , PublisherType.APP_SERVICE, CategoryType.MALICIOUS, EventType.INVALID_INPUTS, Classifier.FAILURE }`

If it's outside the scope of the current release, we should at least create a story to track this.

service/pom.xml
```diff
@@ -41,6 +41,12 @@
         </dependencies>
     </dependencyManagement>
     <dependencies>
+    
+        <dependency>
```
anubhavi25: This isn't necessary since the `audit-sdk` dependencies have already been added to the POMs. All that's necessary is to change the version in the `<audit-sdk.version>0.0.x-SNAPSHOT</audit-sdk.version>` line in the top-level POM.

That said, we should add something like the following to the `public` Maven profile in this POM to exclude compilation of all `src` and `test` classes in the `com.ge.predix.acs.audit` package:

```
            <build>
                <plugins>
                    <plugin>
                        <groupId>org.apache.maven.plugins</groupId>
                        <artifactId>maven-compiler-plugin</artifactId>
                        <version>3.3</version>
                        <configuration>
                            <excludes>
                                <exclude>com/ge/predix/acs/audit/*</exclude>
                            </excludes>
                            <testExcludes>
                                <testExclude>com/ge/predix/acs/audit/*</testExclude>
                            </testExcludes>
                        </configuration>
                    </plugin>
                </plugins>
            </build>
```

service/src/main/java/com/ge/predix/acs/audit/PredixEventMapper.java
```diff
@@ -0,0 +1,44 @@
...
+import com.ge.predix.audit.sdk.message.AuditEventV2.AuditEventV2Builder;
+
+public class PredixEventMapper {
+
+    public AuditEventV2 map(final com.ge.predix.audit.AuditEvent auditEvent) {
+        Map<String, String> auditPayload = new HashMap<String, String>();
+        auditPayload.put("httpMethod", auditEvent.getMethod());
+        auditPayload.put("requestBody", auditEvent.getRequestBody());
+        auditPayload.put("responseBody", auditEvent.getResponseBody());
+
```
anurag-gujral: Should we create a json string out of auditPayload and then store it in payload field.

service/src/test/java/com/ge/predix/acs/audit/PredixEventProcessorTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+import org.testng.Assert;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import com.ge.predix.audit.sdk.AuditClient;
+import com.ge.predix.audit.sdk.exception.AuditException;
+import com.ge.predix.audit.sdk.message.AuditEventV2;
+import com.ge.predix.eventhub.EventHubClientException;
```
anubhavi25: To allow this import to be resolved, we probably want to explicitly add the EventHub SDK dependency to the `predix` Maven profile of `acs/service/pom.xml`:

```
                <dependency>
                    <groupId>com.ge.predix.eventhub</groupId>
                    <artifactId>predix-event-hub-sdk</artifactId>
                    <version>${predix-event-hub-sdk.version}</version>
                </dependency>
```

service/src/test/java/com/ge/predix/acs/audit/PredixEventProcessorTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Mock
+    private AuditClient mockedClient;
+
+    @Mock
+    private PredixEventMapper mockedMapper;
+
+    @BeforeMethod
+    public void setup() throws AuditException, EventHubClientException {
+        this.eventProcessor = new PredixEventProcessor();
```
anubhavi25: Contrary to what we thought, this line to instantiate `this.eventProcessor` doesn't seem necessary (runs without a `NullPointerException` for me)

service/src/test/java/com/ge/predix/acs/audit/PredixEventProcessorTest.java
```diff
@@ -0,0 +1,45 @@
...
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void testPredixEventProcess() throws AuditException, EventHubClientException {
+        AuditEventV2 mockPredixEvent = Mockito.mock(AuditEventV2.class);
```
anubhavi25: Slightly better since it passes a mock object to `process(...)` instead of `null`:

```
    @Test
    public void testPredixEventProcess() throws AuditException, EventHubClientException {
        AuditEvent mockAuditEvent = Mockito.mock(AuditEvent.class);
        AuditEventV2 mockPredixEvent = Mockito.mock(AuditEventV2.class);
        Mockito.when(mockedMapper.map(mockAuditEvent)).thenReturn(mockPredixEvent);

        Assert.assertTrue(this.eventProcessor.process(mockAuditEvent));
        Mockito.verify(mockedClient).audit(mockPredixEvent);
    }
```

anubhavi25 (changes requested):
As discussed, I checked in the CheckStyle fixes to `service/src/test/java/com/ge/predix/acs/audit/PredixEventMapperTest.java` so please pull those changes.

service/src/test/java/com/ge/predix/acs/audit/PredixEventMapperTest.java
FrankGasparovic: Added story for this in audit feature

service/src/test/java/com/ge/predix/acs/audit/PredixEventProcessorTest.java
FrankGasparovic: Removed

service/src/test/java/com/ge/predix/acs/audit/PredixEventProcessorTest.java
FrankGasparovic: Added

service/src/test/java/com/ge/predix/acs/audit/PredixEventProcessorTest.java
FrankGasparovic: added

service/src/test/java/com/ge/predix/acs/audit/PredixEventMapperTest.java
FrankGasparovic: Added

service/src/main/java/com/ge/predix/acs/audit/PredixEventProcessor.java
FrankGasparovic: Removed

service/src/main/java/com/ge/predix/acs/audit/PredixEventProcessor.java
FrankGasparovic: Added

service/src/main/java/com/ge/predix/acs/audit/PredixEventMapper.java
FrankGasparovic: Removed

service/src/main/java/com/ge/predix/acs/audit/PredixEventMapper.java
FrankGasparovic: Am now serializing

service/src/main/java/com/ge/predix/acs/audit/PredixEventMapper.java
FrankGasparovic: Removed

anubhavi25 (approved):

service/src/main/java/com/ge/predix/acs/audit/PredixEventProcessor.java
anurag-gujral: This has not been addressed, frank can you please comment?

service/src/test/java/com/ge/predix/acs/audit/PredixEventMapperTest.java
anurag-gujral: The story for CategoryType.MALICIOUS /input validation has been added to the backlog.

service/src/main/java/com/ge/predix/acs/audit/PredixEventProcessor.java
anubhavi25: @anurag-gujral, @FrankGasparovic has addressed this above.

service/src/main/java/com/ge/predix/acs/audit/PredixEventMapper.java
```diff
@@ -0,0 +1,65 @@
...
+    private ObjectMapper objectMapper;
+
+    public PredixEventMapper() {
+        this.objectMapper = new ObjectMapper();
+    }
+
+    public AuditEventV2 map(final AuditEvent auditEvent) {
+        Map<String, String> auditPayload = new HashMap<>();
+        auditPayload.put("httpMethod", auditEvent.getMethod());
```
anurag-gujral: May next time when we commit we can make hardcoded string to private final String members

service/src/main/java/com/ge/predix/acs/audit/PredixEventMapper.java
```diff
@@ -0,0 +1,65 @@
...
+        auditPayload.put("requestBody", auditEvent.getRequestBody());
+        auditPayload.put("responseBody", auditEvent.getResponseBody());
+
+        String payload = "";
+        try {
+            payload = this.objectMapper.writeValueAsString(auditPayload);
+        } catch (JsonProcessingException e) {
```
anurag-gujral: This does not look correct, the map function should throw an exception instead of just logging.

service/src/main/java/com/ge/predix/acs/audit/PredixEventProcessor.java
```diff
@@ -0,0 +1,34 @@
...
+    @Autowired
+    private AuditClient auditClient;
+
+    @Autowired
+    private PredixEventMapper eventMapper;
+
+    @Override
+    public boolean process(final AuditEvent auditEvent) {
+        AuditEventV2 predixEvent = eventMapper.map(auditEvent);
+        try {
```
anurag-gujral, line 25: Need proper exception handling rather than catching the exception and returning boolean

service/src/main/java/com/ge/predix/acs/audit/PredixEventMapper.java
FrankGasparovic: Only used in one place at the moment.

service/src/main/java/com/ge/predix/acs/audit/PredixEventMapper.java
FrankGasparovic: We are only constructing well formed JSON in this map method with potentially "garbage" as values in the JSON. I don't see a reason that we should throw this exception.

service/src/main/java/com/ge/predix/acs/audit/PredixEventProcessor.java
FrankGasparovic: Boolean is required because of the way we created the tests for the AuditEventProcessor. Also we shouldn't add the throws at the interface level because different implementations of AuditEventProcessor may not use the Predix-Audit SDK.

service/src/main/java/com/ge/predix/acs/audit/PredixEventMapper.java
FrankGasparovic: Also the values in the auditPayload cannot be null, so we shouldn't have problems.

PR #94, opened 2017-03-02
US77483: Added a class to serialize/deserialize the response body fro…

anubhavi25:
…m adapters

### DIFF
--- a/model/src/main/java/com/ge/predix/acs/rest/attribute/adapter/AttributesResponse.java
+++ b/model/src/main/java/com/ge/predix/acs/rest/attribute/adapter/AttributesResponse.java
@@ -1,0 +1,36 @@
+package com.ge.predix.acs.rest.attribute.adapter;
+
+import java.util.Set;
+
+import com.ge.predix.acs.model.Attribute;
+
+public class AttributesResponse {
+
+    private Set<Attribute> attributes;
+    private String id;
+
+    public AttributesResponse() {
+        // Default constructor necessary for Jackson
+    }
+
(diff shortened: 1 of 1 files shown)


model/src/main/java/com/ge/predix/acs/model/AdapterResponse.java
```diff
@@ -0,0 +1,34 @@
+package com.ge.predix.acs.model;
```
sanjeevchopra: `com.ge.predix.acs.rest.attribute.adapter`

model/src/main/java/com/ge/predix/acs/model/AdapterResponse.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.Set;
+
+public class AdapterResponse {
```
sanjeevchopra: `AttributesResponse` will allow for other response types from adapters in future

sanjeevchopra (dismissed):

model/src/main/java/com/ge/predix/acs/model/AdapterResponse.java
anubhavi25: Done

model/src/main/java/com/ge/predix/acs/model/AdapterResponse.java
anubhavi25: Done

anubhavi25:
@sanjeevchopra please do one final review so I can squash changes

PR #96, opened 2017-03-10
 US65642: log AuditEvents to predix audit service

FrankGasparovic:


model/src/main/java/com/ge/predix/audit/rest/PredixAuditRequest.java
```diff
@@ -0,0 +1,49 @@
+package com.ge.predix.audit.rest;
+
+public class PredixAuditRequest {
```
sanjeevchopra: Please move this, response and AuditRecord classes to integration-test project if they are not used by the service.

sanjeevchopra (commented):
Didnt have a chance to review the whole change, I have one comment though.
@anubhavi25  please review the PR

service/src/main/java/com/ge/predix/acs/audit/ACSAuditConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+            sdkConfig = vcapLoaderService.getConfigFromVcap();
+            sdkConfig.setClientType(AuditClientType.ASYNC);
+            sdkConfig.setUaaUrl(auditUaaUrl);
+            sdkConfig.setUaaClientId(auditUaaClientId);
+            sdkConfig.setUaaClientSecret(auditUaaClientSecret);
+
+        } catch (VcapLoadException e) {
+            e.printStackTrace();
+        }
+        AuditClient auditClient = new AuditClient(sdkConfig, auditCallback());
```
anurag-gujral: This does not look right createAuditClient needs to throw an except when it fails but here we are catching the vLoadException and still initializing the auditClient

service/src/main/java/com/ge/predix/acs/audit/ACSAuditConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+        return auditClient;
+    }
+
+    public AuditCallback auditCallback() {
```
anurag-gujral, line 58: If AuditCallBack class is only use for integration tests then we should move to IntegrationTest project as mentioned by Sanjeev earlier.

service/src/main/java/com/ge/predix/acs/audit/PredixEventProcessor.java
```diff
@@ -1,34 +1,63 @@
...
+@Component
+@Profile("predixAudit")
 public class PredixEventProcessor implements AuditEventProcessor {
 
     private static final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(PredixEventProcessor.class);
 
-    @Autowired
-    private AuditClient auditClient;
+    // Remove this and use configuration eventually
+    private static final Set<String> EXCLUSIONS = Collections.singleton("/connector");
```
anubhavi25: Shouldn't this be `/v1/connector` since our APIs are prefixed with the version? Also, once https://github.com/predix/acs/pull/103 is in develop, this should change to `Collections.singleton(AcsApiUriTemplates.V1 + AcsApiUriTemplates.CONNECTOR_URL)` to prevent hardcoding the URI path.

service/src/main/java/com/ge/predix/acs/audit/ACSAuditConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+import com.ge.predix.audit.sdk.validator.ValidatorReport;
+import com.ge.predix.eventhub.EventHubClientException;
+
+@Configuration
+@Profile("predixAudit")
+public class ACSAuditConfiguration {
+
+    private static final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(ACSAuditConfiguration.class);
+
+    private VcapLoaderServiceImpl vcapLoaderService = new VcapLoaderServiceImpl();
```
anubhavi25: Make this `static final` since we should only need to do this once?

service/src/main/java/com/ge/predix/acs/audit/PredixEventProcessor.java
```diff
@@ -1,34 +1,63 @@
...
     @Autowired
     private PredixEventMapper eventMapper;
 
+    @Autowired
+    private AuditClient auditClient;
+
     @Override
     public boolean process(final AuditEvent auditEvent) {
-        AuditEventV2 predixEvent = eventMapper.map(auditEvent);
+        if (null == this.auditClient || isExclusion(auditEvent)) {
```
anubhavi25: The `null == this.auditClient` condition is redundant since it will always be false (since the `predixAudit` Spring profile always wires in the `auditClient` bean).

service/src/main/java/com/ge/predix/acs/audit/ACSAuditConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private AuditConfiguration sdkConfig;
```
anubhavi25: Field can be converted to a local variable in the `auditClient` bean below (since it's only used there)

acs-integration-tests/pom.xml
```diff
@@ -382,7 +385,11 @@
                             <memory>1024</memory>
                             <buildpack>java-buildpack</buildpack>
                             <env>
-                                <SPRING_PROFILES_ACTIVE>integration,predix,cloud-redis,cloudDbConfig</SPRING_PROFILES_ACTIVE>
+                                <SPRING_PROFILES_ACTIVE>integration,predix,cloud-redis,cloudDbConfig,predixAudit</SPRING_PROFILES_ACTIVE>
```
anubhavi25: We don't seem to be using the `predixAudit` Spring profile anywhere in integration tests so remove it so there's one less reference to remove in a later refactoring?

acs-integration-tests/pom.xml
```diff
@@ -538,13 +553,18 @@
                             <memory>1024</memory>
                             <buildpack>java-buildpack</buildpack>
                             <env>
-                                <SPRING_PROFILES_ACTIVE>integration,predix,cloud-redis,titan,cloudDbConfig</SPRING_PROFILES_ACTIVE>
+                                <SPRING_PROFILES_ACTIVE>integration,predix,cloud-redis,titan,cloudDbConfig,predixAudit</SPRING_PROFILES_ACTIVE>
```
anubhavi25: See comment above re. removing `predixAudit`

service/src/test/java/com/ge/predix/acs/audit/PredixEventProcessorTest.java
```diff
@@ -35,11 +35,26 @@ public void setup() throws AuditException, EventHubClientException {
...
 
         Assert.assertTrue(this.eventProcessor.process(mockAuditEvent));
-        Mockito.verify(mockedClient).audit(mockPredixEvent);
+        Mockito.verify(this.mockedClient).audit(mockPredixEvent);
+    }
+
+    @Test
+    public void testControllerExclusions() {
+        AuditEvent auditEvent = Mockito.mock(AuditEvent.class);
+        Mockito.doReturn("http://acs.com/connector/test123").when(auditEvent).getRequestUri();
```
anubhavi25: If you make the change above to use `/v1/connector`, change this test as well

anubhavi25 (dismissed):

PR #97, opened 2017-03-11
US92112: Encrypt and Store adapter client-secret in ACS db

FrankGasparovic:


service/src/main/java/com/ge/predix/acs/encryption/AcsEncryption.java
```diff
@@ -0,0 +1,69 @@
...
+            SecretKeySpec skeySpec = new SecretKeySpec(key.getBytes("UTF-8"), "AES");
+
+            Cipher cipher = Cipher.getInstance(ALGO);
+            cipher.init(Cipher.ENCRYPT_MODE, skeySpec, iv);
+
+            byte[] encrypted = cipher.doFinal(value.getBytes());
+            byte[] result = Bytes.concat(ivBytes, encrypted);
+
+            return Base64.encodeBase64String(result);
+        } catch (Exception ex) {
```
sanjeevchopra: * catch `Throwable` to be safe
* pls remove the printStackTrace
* log a error (excluding the encryption key)

service/src/main/java/com/ge/predix/acs/encryption/AcsEncryption.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private static final String ALGO = "AES/CBC/PKCS5PADDING";
+
+    public AcsEncryption(final String key) {
+        this.key = key;
+    }
+
+    public String encrypt(final String value) {
+        try {
+            if (this.key.length() != 24) {
```
sanjeevchopra: are there any other requirements on the key ?

service/src/main/java/com/ge/predix/acs/config/EncryptionConfig.java
```diff
@@ -0,0 +1,20 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+import com.ge.predix.acs.encryption.AcsEncryption;
+
+@Configuration
+public class EncryptionConfig {
+
+    @Value("${ENCRYPTION_KEY}")
```
sanjeevchopra: wondering if we should use a obscure name for obfuscation

service/src/main/java/com/ge/predix/acs/encryption/AcsEncryption.java
```diff
@@ -0,0 +1,69 @@
...
+            byte[] result = Bytes.concat(ivBytes, encrypted);
+
+            return Base64.encodeBase64String(result);
+        } catch (Exception ex) {
+            ex.printStackTrace();
+            throw new RuntimeException("Unable to encrypt" + value);
+        }
+    }
+
+    public String decrypt(final String encrypted) {
```
sanjeevchopra: comments from encrypt apply here as well

service/src/main/java/com/ge/predix/acs/encryption/AcsEncryption.java
```diff
@@ -0,0 +1,69 @@
...
+            if (this.key.length() != 24) {
+                throw new RuntimeException("Encryption key must be string of length 24");
+            }
+            byte[] ivBytes = new byte[16];
+            SecureRandom.getInstanceStrong().nextBytes(ivBytes);
+            IvParameterSpec iv = new IvParameterSpec(ivBytes);
+            SecretKeySpec skeySpec = new SecretKeySpec(key.getBytes("UTF-8"), "AES");
+
+            Cipher cipher = Cipher.getInstance(ALGO);
+            cipher.init(Cipher.ENCRYPT_MODE, skeySpec, iv);
```
sanjeevchopra: can the key and cipher setup be done once in the constructor ?

service/src/main/java/com/ge/predix/acs/encryption/AcsEncryption.java
```diff
@@ -0,0 +1,69 @@
...
+
+import javax.crypto.Cipher;
+import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+
+import org.apache.commons.codec.binary.Base64;
+
+import com.google.common.primitives.Bytes;
+
+public class AcsEncryption {
```
sanjeevchopra: `AESEncrypter` ?

sanjeevchopra (changes requested):
can you rebase from the branch which has the ConnectorService, so you can invoke the encryption/decryption ? or we can wait for that branch to merge to develop.

service/src/main/java/com/ge/predix/acs/encryption/AcsEncryption.java
FrankGasparovic: Went with Encryptor

service/src/main/java/com/ge/predix/acs/encryption/AcsEncryption.java
FrankGasparovic: Not that I know of. We are doing 192 bit encryption, the key just needs to have that number of bits and we should be good to go.

service/src/main/java/com/ge/predix/acs/encryption/AcsEncryption.java
FrankGasparovic: Good call. Moved.

service/src/main/java/com/ge/predix/acs/encryption/AcsEncryption.java
FrankGasparovic: Created custom exceptions based on the error and added log statements.

service/src/main/java/com/ge/predix/acs/encryption/AcsEncryption.java
FrankGasparovic: Changed.

service/src/main/java/com/ge/predix/acs/config/EncryptionConfig.java
FrankGasparovic: Obfuscation from who? If someone is able to see our environment variables in cf, they are already a privileged user.

service/src/main/java/com/ge/predix/acs/encryption/Encryptor.java
```diff
@@ -0,0 +1,9 @@
+package com.ge.predix.acs.encryption;
+
+public interface Encryptor {
```
sanjeevchopra: To make this generally usable, we could build this as a credential store interface. It has 2  functions storage and encryption/decryption.  This service could be used as-is with audit or any other credential which ACS needs to store.

`public interface CredentialStore {

    String encryptAndStore(String id, String key, String value);

    String findAndDecrypt(String id, String key, String encrypted);

}`

* id - will be used for lookup and storage
* key -  for encryption/decryption OR authN token in case of vault)

sanjeevchopra: discussed with Frank. Doesnt quite work because when using a store, you dont really have a 'encrypted value'.  Probably best as 2 separate services - Encryptor and SecureStorage.  You only need one or the other   (Encryptor + regular storage  OR SecureStorage)

service/src/main/java/com/ge/predix/acs/attribute/connector/management/AttributeConnectorServiceImpl.java
```diff
@@ -29,15 +36,22 @@ public boolean upsertResourceConnector(final AttributeConnector connector) {
 
         AttributeConnectorEntity connectorEntity = zoneEntity.getResourceAttributeConnector();
         boolean isCreated = null == connectorEntity;
-        
+
+        Set<AttributeAdapterConnection> attributeAdapterConnections = new HashSet<>();
+        connector.getAdapters().parallelStream().forEach(a -> {
+            a.setUaaClientSecret(this.encryptor.encrypt(a.getUaaClientSecret()));
+            attributeAdapterConnections.add(a);
```
sanjeevchopra: * why do you need a new set of connections ? the secret is already updated by reference
* pls make sure HashSet indeed supports parallel streams
* code will read better if you rename `a` to `adapterConnection`

service/src/main/java/com/ge/predix/acs/attribute/connector/management/AttributeConnectorServiceImpl.java
```diff
@@ -46,7 +60,16 @@ public boolean upsertResourceConnector(final AttributeConnector connector) {
     @Override
     public AttributeConnector retrieveResourceConnector() {
         ZoneEntity zoneEntity = this.zoneResolver.getZoneEntityOrFail();
-        return this.connectorConverter.toConnector(zoneEntity.getResourceAttributeConnector());
+        AttributeConnector connector = this.connectorConverter.toConnector(zoneEntity.getResourceAttributeConnector());
+        if (null != connector) {
+            Set<AttributeAdapterConnection> attributeAdapterConnections = new HashSet<>();
```
sanjeevchopra: same comments as upsert

service/src/main/java/com/ge/predix/acs/attribute/connector/management/dao/AttributeAdapterConnectionEntity.java
```diff
@@ -32,7 +31,7 @@
     @Column(name = "adapter_client_id", nullable = false, length = 128)
     private String adapterClientId;
 
-    @Transient
+    @Column(name = "adapter_client_secret", nullable = false, length = 128)
```
sanjeevchopra: does the encryption algo produce a value of the same length ?  Just making sure if 128 is the appropriate lenght

service/src/main/java/com/ge/predix/acs/config/EncryptionConfig.java
```diff
@@ -0,0 +1,16 @@
+package com.ge.predix.acs.config;
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+import com.ge.predix.acs.encryption.Encryptor;
+
+@Configuration
+public class EncryptionConfig {
```
sanjeevchopra: Please remove this class and make Encryptor a spring bean.
* Also AttributeAdatperConnectionEntity should have a transient getUaaClientSecret() which returns the decrypted secret
* The persistent field can be renamed to something like getEncryptedUaaClientSecret()    (the whole entity is changing to transient now based on our json change)

service/src/main/java/com/ge/predix/acs/encryption/Encryptor.java
```diff
@@ -0,0 +1,94 @@
...
+        try {
+            byte[] ivBytes = new byte[IV_LENGTH_IN_BYTES];
+            SecureRandom.getInstanceStrong().nextBytes(ivBytes);
+            IvParameterSpec iv = new IvParameterSpec(ivBytes);
+            SecretKeySpec skeySpec = new SecretKeySpec(this.encryptionKey.getBytes(ENCODING), ALGO);
+
+            this.cipher.init(Cipher.ENCRYPT_MODE, skeySpec, iv);
+
+            byte[] encrypted = this.cipher.doFinal(value.getBytes());
+            byte[] result = Bytes.concat(ivBytes, encrypted);
```
sanjeevchopra, line 48: there is limited value to concat'ing a salt here, given we our code is open source.   If we keep it, pls add some comments to explain the steps here.

service/src/test/java/com/ge/predix/acs/encryption/EncryptorTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class EncryptorTest {
+
+    private static final String ENCRYPTION_KEY_VAR_NAME = "encryptionKey";
+    private static final String VALUE_TO_ENCRYPT = "testValue";
+
+    @Test
+    public void testEncryptCompleteFlow() {
+        Encryptor encryption = Encryptor.getInstance();
+        ReflectionTestUtils.setField(encryption, ENCRYPTION_KEY_VAR_NAME, "FooBarFooBarFooB");
```
sanjeevchopra: Assert.asertNotEquals(encryption.encrypt(VALUE_TO_ENCRYPT), VALUE_TO_ENCRYPT)

service/src/main/java/com/ge/predix/acs/encryption/Encryptor.java
irinaepshteyn: IV is required by CBC cipher mode.  We could switch to using ECB instead and then IV will not be required.  ECB is considered weaker than CBC since the same blocks of plain text encrypted with ECB result in the same encrypted blocks.

sanjeevchopra: Frank, explained that this is useful to prevent against dictionary attacks. Lets keep it and add some comments here 👍

service/src/main/java/com/ge/predix/acs/encryption/Encryptor.java
```diff
@@ -0,0 +1,98 @@
...
+import org.springframework.stereotype.Component;
+
+import com.google.common.primitives.Bytes;
+
+@Component
+public final class Encryptor {
+
+    private String encryptionKey;
+
+    private static final Encryptor INSTANCE = new Encryptor();
```
sanjeevchopra: I don't think this will work, because the object instance being created here will not be spring managed, and hence not receive the ENCRYPTION_KEY value from the env var.  Is there a integration test for this yet ?
* On 2nd thought the cleanest way to solve this would be to have AttributeConnectorService decrypt the client secret. Since it is spring managed, it can autowire the Encryptor to it.  You could do the json serialization/des-ser there too.

sanjeevchopra (changes requested):

sanjeevchopra:
I also pushed a commit with some refactoring on how to handle json serialization/de-ser' .  If it looks better - please keep it. Tests are clean with this change.

Also - if you move the ser/de-ser to the service, the refactoring still applies - should be easily portable.

service/src/main/java/com/ge/predix/acs/encryption/Encryptor.java
sanjeevchopra: I pushed a commit to make this change.

sanjeevchopra:
This branch is ready to merge, imo.

sanjeevchopra:
PR build passed for predix-cloud - https://predix1.jenkins.build.ge.com/job/Predix-Security/job/ACS/job/ACS-PullRequest/job/acs-integration-pullrequest/291/

(issue seems to be postgres resources - it cannot create 4 db instances at a time.  I had to remove some db instances to allow the above build to go through and remove 'cf login' command from the jenkins job)

Filed https://gesoftware.service-now.com/nav_to.do?uri=incident.do?sys_id=3bc1cd164f25f600b3d87acd0210c7d5 for this

PR #99, opened 2017-03-13
Aggregate attributes from remote adapters

sanjeevchopra:


acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationWithAttributeConnectorIT.java
```diff
@@ -0,0 +1,162 @@
...
+        httpHeaders.set(PolicyHelper.PREDIX_ZONE_ID, this.zone.getSubdomain());
+        return httpHeaders;
+    }
+
+    private void configureAttributeConnector() throws IOException {
+        AttributeConnector attributeConnector = new AttributeConnector();
+        attributeConnector.setIsActive(true);
+        attributeConnector.setMaxCachedIntervalMinutes(60);
+        attributeConnector.setAdapters(new HashSet<>(Collections.singletonList(
+                new AttributeAdapterConnection("https://acs-asset-adapter.run.aws-usw02-dev.ice.predix.io/v1/attribute",
```
sanjeevchopra: please parameterize  the adapter settings.  settings will change for each env

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationWithAttributeConnectorIT.java
```diff
@@ -0,0 +1,162 @@
...
+                                + ".io/oauth/token", "connector", "connector"))));
+        this.resourceAttributeConnectorUrl = URI.create(this.zoneHelper.getAcsBaseURL() + AcsApiUriTemplates.V1
+                + AcsApiUriTemplates.RESOURCE_CONNECTOR_URL);
+        HttpHeaders httpHeaders = this.getHeadersWithZoneId();
+        httpHeaders.setContentType(MediaType.APPLICATION_JSON);
+        this.acsAdminRestTemplate.exchange(this.resourceAttributeConnectorUrl, HttpMethod.PUT,
+                new HttpEntity<>(attributeConnector, httpHeaders), AttributeConnector.class);
+    }
+
+    private void setupPredixAcs() throws IOException {
```
sanjeevchopra: can these 3 methods can be shared between different integ tests - maybe in a new version of createTestZone method, which tests can start using over time

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationWithAttributeConnectorIT.java
```diff
@@ -0,0 +1,162 @@
...
+    @BeforeClass
+    void beforeClass() throws IOException {
+        if (Arrays.asList(this.environment.getActiveProfiles()).contains("public")) {
+            setupPublicAcs();
+        } else {
+            setupPredixAcs();
+        }
+
+        this.zone = this.zoneHelper.createTestZone(this.acsAdminRestTemplate, this.acsZone1Name, this.registerWithZac);
+        this.configureAttributeConnector();
```
sanjeevchopra: where is the data in the adapter populated ? please add it as part of this test.

service/src/main/java/com/ge/predix/acs/attribute/readers/ExternalAttributeReader.java
```diff
@@ -29,20 +55,78 @@ public ExternalAttributeReader(final AttributeCache attributeCache) {
...
+        HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory(httpClient);
+        requestFactory.setReadTimeout(this.adapterHttpClientTimeout);
+        requestFactory.setConnectTimeout(this.adapterHttpClientTimeout);
+        requestFactory.setConnectionRequestTimeout(this.adapterHttpClientTimeout);
+        restTemplate.setRequestFactory(requestFactory);
+    }
+
+    private OAuth2RestTemplate getAdapterOauth2RestTemplate(
+            final AttributeAdapterConnectionEntity attributeAdapterConnectionEntity) {
+        ClientCredentialsResourceDetails clientCredentials = new ClientCredentialsResourceDetails();
```
sanjeevchopra: rest template should be cached in this object, instead of creating on every request. This will prevent a "get token" on every request
* pls review anything else which can be done once or at construction time

service/src/main/java/com/ge/predix/acs/attribute/readers/ExternalAttributeReader.java
```diff
@@ -29,20 +55,78 @@ public ExternalAttributeReader(final AttributeCache attributeCache) {
...
+            HttpEntity<String> adapterEntity = new HttpEntity<>(headers);
+
+            String adapterUrl = UriComponentsBuilder
+                    .fromUriString(attributeAdapterConnectionEntity.getAdapterEndpoint()).queryParam(ID, identifier)
+                    .toUriString();
+
+            ResponseEntity<AttributesResponse> attributesResponse;
+            try {
+                attributesResponse = this.getAttributesFromAdapter(adapterUrl, adapterEntity, adapterRestTemplate);
+            } catch (final RestClientException e) {
```
sanjeevchopra: catch 'Throwable' here to ensure evaluation service can handle the AttributeRetrievalException
* Lets make sure there is a test which asserts that policy eval service handles any exception and returns a INDETERMINATE

service/src/main/java/com/ge/predix/acs/attribute/readers/ExternalAttributeReader.java
```diff
@@ -29,20 +55,78 @@ public ExternalAttributeReader(final AttributeCache attributeCache) {
                 attributes = this.getAttributesFromAdapters(identifier);
                 this.attributeCache.setAttributes(identifier, attributes);
 
-                // TODO: Reset the TTL for the policy evaluation cache key concerning this resource to:
-                // total customer TTL - this.attributeCache.getConfiguredTtlInSeconds()
+                //long policyEvalDecisionCacheTtlInSeconds =
```
sanjeevchopra: Policy eval simply constructs a key and call the decision cache.  When to use the cached decision will have to be in the decision cache layer.  The decision cache does not have a TTL configured on the cache itself - see `AbstractPolicyEvaluationCache.isCachedRequestInvalid` - after looking up both the decision and the lastModifiedTime for the resource(& subject) involved, it decides to use the decision or not.
We need to modify this code path to:
* check if resource or subject connector is enabled
* if yes, the decision cache should be invalidated if its been cached for >= (connector.CacheInterval - attributeCacehInterval - 1 min for safety).   

This is one way to make this functional -  we should reconsider the relationship between decision and attribute cache when we implement a real cache for attributes.  Maybe one abstraction should provide the decision Or attributes (if the decision is stale).

service/src/test/java/com/ge/predix/acs/zone/management/ZoneRepositoryTest.java
```diff
@@ -93,14 +93,6 @@ public void testDeleteConnectorWithCascade() {
         Assert.assertNull(this.adapterRepository.findOne(adapterId));
     }
 
-    @Test
```
sanjeevchopra: why remove ? isnt this the expected behavior right now ?

service/src/test/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationWithAttributeReaderTest.java
```diff
@@ -83,35 +90,47 @@ public void setupMethod() throws Exception {
...
     }
 
+    @SuppressWarnings("unchecked")
+    @Test
+    public void testPolicyEvaluationWhenAdaptersTimeOut() throws Exception {
+        when(this.externalResourceAttributeReader.getAttributes(Mockito.anyString()))
+                .thenThrow(AttributeRetrievalException.class);
+
+        PolicyEvaluationResult evalResult = this.evaluatePolicy();
+        Assert.assertEquals(evalResult.getEffect(), Effect.INDETERMINATE);
```
sanjeevchopra: this test would be better if you do a successful PERMIT first, then only mock the rule above to produce a INDETERMIANTE.   

Also, in this case eval service should set a appropriate message in PoliceEvaluationResult.message indicating some problem with the adapter, along with reference info for the adapter.  Just enough info, so the end-user can pass on the reference to service owner who is using ACS.

service/src/main/java/com/ge/predix/acs/attribute/readers/AttributeReaderFactory.java
```diff
@@ -1,22 +1,48 @@
...
     }
 
     public SubjectAttributeReader getSubjectAttributeReader() {
-        return this.subjectAttributeReader;
+        ZoneEntity zone = this.zoneResolver.getZoneEntityOrFail();
+        if (zone.getSubjectAttributeConnector() == null) {
+            return this.privilegeServiceSubjectAttributeReader;
+        }
+
+        return new ExternalSubjectAttributeReader(zone,
```
sanjeevchopra: This factory needs to make sure that a subject or resource externalreader is only created once for each zone.
* also need to check isActive property on the connector, once its added to the entity

service/src/main/java/com/ge/predix/acs/attribute/readers/ExternalAttributeReader.java
sanjeevchopra: this is not going to work. lets discuss an alternate solution tomorrow.

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationWithAttributeConnectorIT.java
```diff
@@ -0,0 +1,162 @@
...
+import com.ge.predix.test.utils.PolicyHelper;
+import com.ge.predix.test.utils.UaaTestUtil;
+import com.ge.predix.test.utils.ZacTestUtil;
+import com.ge.predix.test.utils.ZoneHelper;
+
+@ContextConfiguration("classpath:integration-test-spring-context.xml")
+public class PolicyEvaluationWithAttributeConnectorIT extends AbstractTestNGSpringContextTests {
+
+    @Value("${ZONE1_NAME:testzone1}")
+    private String acsZone1Name;
```
sanjeevchopra: do you need 3 acs zones for this test ?  Also, pls use a unique zone name to allow concurrent execution  with other integ tests.  (now that we dont use sub-domain routing, any zone name can be provisioned on the fly)

sanjeevchopra:
discussed this with anubhav:
acs-integration test should push the asset adapter and run the connector integ test concurrently with the other integ tests. pls add to a separate test suite, if needed.

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationWithAttributeConnectorIT.java
anubhavi25: Done

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationWithAttributeConnectorIT.java
anubhavi25: I haven't refactored this yet but we can as part of this story or maybe another one?

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationWithAttributeConnectorIT.java
anubhavi25: Done (Asset data is now configured/deconfigured in this test class)

service/src/main/java/com/ge/predix/acs/attribute/readers/ExternalAttributeReader.java
anubhavi25: As discussed, implemented this; a related test hasn't been added yet because it would be time-based and thus error prone.

service/src/main/java/com/ge/predix/acs/attribute/readers/ExternalAttributeReader.java
anubhavi25: Done

service/src/main/java/com/ge/predix/acs/attribute/readers/ExternalAttributeReader.java
anubhavi25: Done. Also there already seems to be a test that asserts that `Effect.INDETERMINATE` is produced when a `RuntimeException` exception is thrown: `PolicyEvaluationServiceTest#testPolicyEvaluationExceptionHandling`

service/src/test/java/com/ge/predix/acs/zone/management/ZoneRepositoryTest.java
anubhavi25: This has been re-introduced.

service/src/test/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationWithAttributeReaderTest.java
anubhavi25: There already is a test for a successful `Effect.PERMIT`: `PolicyEvaluationWithAttributeReaderTest#testPolicyEvaluation` so not sure how adding another one here would add value.

Also the adapter endpoint, UAA token URL and UAA client ID are added as part of `PolicyEvaluationResult.message`.

service/src/main/java/com/ge/predix/acs/attribute/readers/AttributeReaderFactory.java
anubhavi25: Done (uses maps keyed on the zone ID). The `isActive` property is now also checked.

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationWithAttributeConnectorIT.java
anubhavi25: Point taken.. removed all but 1 zone. If we go with a separate integration test suite for connector-related policy evaluation integration tests, we can create a unique zone then but since this currently runs as part of the existing integration test suite I'm reusing a zone.

anubhavi25:
@sanjeevchopra: we should probably create a separate story to push the asset adapter as part of `mvn verify` since I'm not sure how difficult it is to push 2 apps via 2 executions of the `cf-maven-plugin` in the `pre-integration` phase. For now, it's already pushed and available on CF3 integration.

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationWithAttributeConnectorIT.java
```diff
@@ -103,16 +95,16 @@
...
-    private static final String ASSET_URI_PATH_SEGMENT = '/' + IDENTIFIER;
+    private static final String TEST_PART_ID = "part/03f95db1-4255-4265-a509-f7bca3e1fee4";
+    private static final String ASSET_URI_PATH_SEGMENT = '/' + TEST_PART_ID;
 
     private Zone zone;
     private OAuth2RestTemplate acsAdminRestTemplate;
     private OAuth2RestTemplate assetRestTemplate;
     private boolean registerWithZac;
-    private URI resourceAttributeConnectorUrl;
+    private URI resourceAttributeConnectorUrl = URI.create(this.zoneHelper.getAcsBaseURL() + "/v1/resource/connector");
```
anubhavi25: This will throw a `NullPointerException` when the test is run (or not run tests at all) since `this.zoneHelper` isn't initialized yet. We should move this initialization to the `@BeforeClass` method below.

Also the connector URI is `/v1/connector/resource` :) . It might just be better to use `AcsApiUriTemplates.V1 + AcsApiUriTemplates.RESOURCE_CONNECTOR_URL` to construct the URI path like before so that we only have to make a change in one place.

model/src/main/java/com/ge/predix/acs/rest/AttributeConnector.java
```diff
@@ -2,9 +2,12 @@
 
 import java.util.Set;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 public class AttributeConnector {
     private boolean isActive;
     
+    @JsonProperty(defaultValue="480", required=false)
```
anubhavi25: I don't think this will actually work.. Jackson's core databind currently only uses `defaultValue` for documenting the expected default value of the property instead of actually enforcing it. If you want to keep this, we need to also change the actual assignment to use a default value like: `private int maxCachedIntervalMinutes = 480;`

anubhavi25 (changes requested):

acs-integration-tests/src/test/java/com/ge/predix/integration/test/PolicyEvaluationWithAttributeConnectorIT.java
```diff
@@ -225,30 +213,25 @@ void afterClass() throws IOException {
     }
 
     @Test(dataProvider = "adapterStatusesAndResultingEffects")
-    public void testPolicyEvaluationWithAdapters(final boolean adapterActive, final Effect effect) throws Exception {
-        HttpHeaders httpHeaders = this.getHeadersWithAcsZoneId();
```
anubhavi25: Since the zone ID doesn't change throughout the test, it's more efficient to instantiate the zone headers once and use them instead of constructing multiple instances by calling `zoneHeader()` multiple times

sanjeevchopra: since this is a test, I gave precedence readability over efficiency.

anubhavi25: Can't we get both with: `HttpHeaders zoneHeaders = this.zoneHeader();` ?

service/src/main/java/com/ge/predix/acs/attribute/readers/ExternalAttributeReader.java
```diff
@@ -28,21 +57,77 @@ public ExternalAttributeReader(final AttributeCache attributeCache) {
...
+        restTemplate.setRequestFactory(requestFactory);
+    }
+
+    private OAuth2RestTemplate getAdapterOauth2RestTemplate(
+            final AttributeAdapterConnection attributeAdapterConnectionEntity) {
+        String uaaTokenUrl = attributeAdapterConnectionEntity.getUaaTokenUrl();
+        String uaaClientId = attributeAdapterConnectionEntity.getUaaClientId();
+        String uaaClientSecret = attributeAdapterConnectionEntity.getUaaClientSecret();
+
+        Integer key = new HashCodeBuilder().append(uaaTokenUrl).append(uaaClientId).append(uaaClientSecret)
```
sanjeevchopra: better to implement hashcode in AdapterConnection class and use the object itself as the key, so the key can be leveraged anywhere. plus its better encapsulation

sanjeevchopra:
Pull request build is clean - integration-tests # 318-321
https://predix1.jenkins.build.ge.com/job/Predix-Security/job/ACS/job/ACS-PullRequest/job/acs-integration-pullrequest/320/

PR #100, opened 2017-03-16
US97526: Change attribute connector persistence to JSON

FrankGasparovic:
Signed-off-by: Irina <irina.epshteyn1@ge.com>


service/src/main/java/com/ge/predix/acs/attribute/connector/management/AttributeConnectorServiceImpl.java
```diff
@@ -20,19 +18,17 @@
...
 
-        AttributeConnectorEntity connectorEntity = zoneEntity.getResourceAttributeConnector();
-        boolean isCreated = null == connectorEntity;
-        
-        connectorEntity = connectorConverter.toConnectorEntity(connector);
-        zoneEntity.setResourceAttributeConnector(connectorEntity);
+        boolean isCreated = false;
         try {
+            AttributeConnector existingConnector = zoneEntity.getResourceAttributeConnector();
+            isCreated = (null == existingConnector);
```
anurag-gujral, line 29: Line 29: does not look right to me, if existingConnector is not null then isCreated will be false.

service/src/main/java/com/ge/predix/acs/attribute/connector/management/AttributeConnectorServiceImpl.java
```diff
@@ -46,18 +42,25 @@ public boolean upsertResourceConnector(final AttributeConnector connector) {
     @Override
     public AttributeConnector retrieveResourceConnector() {
         ZoneEntity zoneEntity = this.zoneResolver.getZoneEntityOrFail();
-        return this.connectorConverter.toConnector(zoneEntity.getResourceAttributeConnector());
+        try {
```
anurag-gujral, line 45: should we add throws clause to this function

anurag-gujral: Guys, 
This is good we are using Json thanks for fast turnaround but should we define the fields/format of the Json like:
{
  "CLIENTD" : "ddd",
  "CLIENDSECRET" : " secret",
  "UAAURL" : "uaa_url",
  "endpoint" : "ep"
}

FrankGasparovic:
Build pass [Build Status](https://predix1.jenkins.build.ge.com/job/Predix-Security/job/ACS/job/ACS-PullRequest/job/acs-build-pullrequest/82/)

service/src/main/java/com/ge/predix/acs/attribute/connector/management/AttributeConnectorServiceImpl.java
FrankGasparovic: Yes, this just means that the connector has been updated. If existingConnector is null then isCreated will be true, meaning you are creating a new connector.

service/src/main/java/com/ge/predix/acs/attribute/connector/management/AttributeConnectorServiceImpl.java
FrankGasparovic: My understanding is that is the current format of the Json. The Object Mapper deserializes the object into those fields.

anurag-gujral: Talked with Frank the last comment is not an issue

service/src/main/java/com/ge/predix/acs/zone/management/dao/ZoneEntity.java
```diff
@@ -155,19 +152,27 @@ public String toString() {
...
     }
 
-    public AttributeConnectorEntity getResourceAttributeConnector() {
-        return this.resourceAttributeConnector;
+    public AttributeConnector getResourceAttributeConnector()
+            throws JsonParseException, JsonMappingException, IOException {
+        if (null == this.resourceAttributeConnector) {
+            return null;
+        }
+        return OBJECT_MAPPER.readValue(this.resourceAttributeConnector, AttributeConnector.class);
```
sanjeevchopra, line 164: this method should deserialize the json only when the json has changed 
* keep a separate field of type AttributeConnector
* initialize from json when getter is called, only if null
* setter should also set this field, and serialize to json only if new value  does not equal current value in object

PR #106, opened 2017-03-24
US102589: Require HTTPS for adapter endpoints and UAA token URLs

FrankGasparovic:


service/src/main/java/com/ge/predix/acs/attribute/connector/management/AttributeConnectorServiceImpl.java
```diff
@@ -176,9 +176,15 @@ private void validateAdapterEntityOrFail(final AttributeAdapterConnection adapte
         if (adapter.getAdapterEndpoint() == null || adapter.getAdapterEndpoint().isEmpty()) {
             throw new AttributeConnectorException("Attribute adapter configuration requires a nonempty endpoint URL");
         }
+        if (!adapter.getAdapterEndpoint().startsWith("https://")) {
```
anurag-gujral: Let us check for case insensitiveness for https

service/src/main/java/com/ge/predix/acs/attribute/connector/management/AttributeConnectorServiceImpl.java
```diff
@@ -176,9 +180,23 @@ private void validateAdapterEntityOrFail(final AttributeAdapterConnection adapte
         if (adapter.getAdapterEndpoint() == null || adapter.getAdapterEndpoint().isEmpty()) {
             throw new AttributeConnectorException("Attribute adapter configuration requires a nonempty endpoint URL");
         }
+        try {
+            if (!new URI(adapter.getAdapterEndpoint()).getScheme().equalsIgnoreCase(HTTPS)) {
```
anurag-gujral: I think we should replace this line (new URI(adapter.getAdapterEndpoint()).getScheme().equalsIgnoreCase(HTTPS)) with function in Utils with can be used in multiple places. Also we need to handle the case if scheme is not specified(for example if someone specifies www.abcd.com)

anurag-gujral: The function will be like:
boolean isHttpsUrl(String urlStr) throws  URISyntaxException{
    if( (urlStr != null) && (!urlStr.isEmpty())){
        URI uri = URI(urlStr);
        String schemeUri = url.getScheme();
        if(schemeUri == null || (schemeUri.isEmpty()))
           return false;
          //i dont like the name HTTPS so changing into to HTTPS_STR
       return (HTTPS_STR.equals(schemeUri) == true) ? true : false)
  }

   return false;
}

anubhavi25: Good point regarding missing schemes.. converted to using `java.net.URL` so that if a protocol isn't specified it will throw a `MalformedURLException` (which also means no defensive checks necessary)

anubhavi25:
[Passing PR integration builds: #385-388](https://predix1.jenkins.build.ge.com/job/Predix-Security/job/ACS/job/ACS-PullRequest/job/acs-integration-pullrequest/)

PR #107, opened 2017-03-27
DE25851: Made Encryptor#encrypt thread safe and improved performance …

anubhavi25:
…by using thread local instances of Cipher

Signed-off-by: Anurag <anurag.gujral@ge.com>

### DIFF
--- a/service/src/main/java/com/ge/predix/acs/encryption/Encryptor.java
+++ b/service/src/main/java/com/ge/predix/acs/encryption/Encryptor.java
@@ -1,56 +1,65 @@
 package com.ge.predix.acs.encryption;
 
-import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
 import java.util.Arrays;
 
 import javax.crypto.Cipher;
-import javax.crypto.NoSuchPaddingException;
 import javax.crypto.spec.IvParameterSpec;
 import javax.crypto.spec.SecretKeySpec;
 
 import org.apache.commons.codec.binary.Base64;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import com.google.common.primitives.Bytes;
 
 public final class Encryptor {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(Encryptor.class);
     private static final String ALGO_WITH_PADDING = "AES/CBC/PKCS5PADDING";
     private static final String ALGO = "AES";
-    private static final String ENCODING = "UTF-8";
     private static final int IV_LENGTH_IN_BYTES = 16;
     private static final int KEY_LENGTH_IN_BYTES = 16;
 
-    private Cipher cipher;
-    private String encryptionKey;
-
-    public Encryptor() {
+    private static final ThreadLocal<Cipher> CIPHER_THREAD_LOCAL = ThreadLocal.withInitial(() -> {
         try {
-            this.cipher = Cipher.getInstance(ALGO_WITH_PADDING);
-        } catch (NoSuchAlgorithmException | NoSuchPaddingException e) {
-            LOGGER.error("Can not created instance of cipher with algorithm" + ALGO_WITH_PADDING);
+            return Cipher.getInstance(ALGO_WITH_PADDING);
+        } catch (Throwable e) {
+            throw new CipherInitializationFailureException(
+                    "Could not create instance of cipher with algorithm: " + ALGO_WITH_PADDING, e);
+        }
+    });
+
+    private SecretKeySpec secretKeySpec;
+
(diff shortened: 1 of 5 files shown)


service/src/main/java/com/ge/predix/acs/encryption/Encryptor.java
```diff
@@ -20,31 +22,69 @@
...
         try {
-            this.cipher = Cipher.getInstance(ALGO_WITH_PADDING);
-        } catch (NoSuchAlgorithmException | NoSuchPaddingException e) {
-            LOGGER.error("Can not created instance of cipher with algorithm" + ALGO_WITH_PADDING);
+            CachedCipher cachedCipher = CIPHER_THREAD_LOCAL.get();
+
+            // Rigorous checks to make sure we've absolutely got the correct instance (with correct alg/key/iv/...)
+            if (cachedCipher != null && cachedCipher.isCipherValid(iv, cipherMode, this.secretKeySpec)) {
+                return cachedCipher.cipher;
+            }
```
irinaepshteyn: Caching cipher seems like unnecessary overhead.  isCipherValid does comparison on cipherMode (ENCRYPT_MODE or DECRYPT_MODE) and on IV.  If the same thread is used continuously to encrypt (or decrypt), the cached cipher will be thrown away each time as new IV is generated for every encryption/decryption operation.  If same thread is used to encrypt and then decrypt, cached cypher will be thrown away as mode is different.  I can't come up with the use case when cached cypher is actually used.

service/src/main/java/com/ge/predix/acs/encryption/Encryptor.java
```diff
@@ -20,31 +22,69 @@
...
+
+            Cipher cipher = Cipher.getInstance(ALGO_WITH_PADDING);
+            cipher.init(cipherMode, this.secretKeySpec, new IvParameterSpec(iv));
+            CIPHER_THREAD_LOCAL.set(new CachedCipher(cipherMode, this.secretKeySpec, cipher));
+            return cipher;
+        } catch (NoSuchAlgorithmException | NoSuchPaddingException | InvalidAlgorithmParameterException
+                | InvalidKeyException e) {
```
irinaepshteyn: It's better to catch a Throwable here and wrap anything that could go wrong with CipherInitializationFailureException

service/src/main/java/com/ge/predix/acs/encryption/Encryptor.java
```diff
@@ -77,6 +113,6 @@ public void setEncryptionKey(final String encryptionKey) {
         if (null == encryptionKey || encryptionKey.length() != KEY_LENGTH_IN_BYTES) {
             throw new SymmetricKeyValidationException("Encryption key must be string of length " + KEY_LENGTH_IN_BYTES);
         }
-        this.encryptionKey = encryptionKey;
+        this.secretKeySpec = new SecretKeySpec(encryptionKey.getBytes(), ALGO);
```
irinaepshteyn: We could unload key length validation to [SecretKeySpec(byte[] key, int offset, int len, String algorithm)](https://docs.oracle.com/javase/8/docs/api/index.html?javax/crypto/spec/SecretKeySpec.html) constructor.  It throws IllegalArgumentException if algorithm is null or key is null, empty, or too short, i.e. key.length-offset<len.

irinaepshteyn (dismissed):
Done reviewing.  Please review my comments.

service/src/main/java/com/ge/predix/acs/encryption/Encryptor.java
anubhavi25: Apart from maybe allowing flexibility to change the algorithm and not having to change this code, I agree with you. For now, I've changed it to just cache the `Cipher` instance thread-locally (agnostic of `Cipher#init`).

service/src/main/java/com/ge/predix/acs/encryption/Encryptor.java
anubhavi25: Done

service/src/main/java/com/ge/predix/acs/encryption/Encryptor.java
anubhavi25: Agreed. In the case that the key is longer than 16 bytes, the first 16 will be used (since the `offset` is 0)

anubhavi25:
[Passing PR Integration Builds: #376-379](https://predix1.jenkins.build.ge.com/job/Predix-Security/job/ACS/job/ACS-PullRequest/job/acs-integration-pullrequest/)

irinaepshteyn (approved):
Ready to merge

PR #115, opened 2017-04-10
US66693: Refactor policy evaluation decision caching to allow eviction of keys

FrankGasparovic:
Signed-off-by: Anubhav <anubhav.saggi@ge.com>

### DIFF
--- a/service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
+++ b/service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
@@ -19,12 +19,13 @@ package com.ge.predix.acs.policy.evaluation.cache;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 
 import org.codehaus.jackson.map.ObjectMapper;
@@ -41,6 +42,27 @@ import com.ge.predix.acs.privilege.management.dao.ResourceEntity;
 import com.ge.predix.acs.privilege.management.dao.SubjectEntity;
 import com.ge.predix.acs.rest.PolicyEvaluationResult;
 
+enum EntityType {
+    RESOURCE {
+        @Override
+        public String toString() {
+            return "resource";
+        }
+    },
+    SUBJECT {
+        @Override
+        public String toString() {
+            return "subject";
+        }
+    },
+    POLICY_SET {
+        @Override
+        public String toString() {
+            return "policy set";
+        }
+    }
+}
+
 @Component
 public abstract class AbstractPolicyEvaluationCache implements PolicyEvaluationCache {
 
@@ -50,6 +72,16 @@ public abstract class AbstractPolicyEvaluationCache implements PolicyEvaluationC
     private static final Logger LOGGER = LoggerFactory.getLogger(AbstractPolicyEvaluationCache.class);
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
+    /**
+     * This method will get the Policy Evaluation Result in the cache. It will check if any of the subject, policy
+     * sets or resolved resource URI's have a timestamp in the cache after the timestamp of the Policy Evaluation
+     * Result. If the key is not in the cache or the result is invalidated, it will return null. Also it will remove
+     * the Policy EvaluationResult so that subsequent evaluations won't find the key in the cache.
+     *
+     * @param key
+     *            The Policy Evaluation key to retrieve.
+     * @return The Policy Evaluation Result if the key is the the cache and the result isn't invalidated, or null
+     */
     @Override
     public PolicyEvaluationResult get(final PolicyEvaluationRequestCacheKey key) {
         String redisKey = key.toRedisKey();
@@ -67,6 +99,19 @@ public abstract class AbstractPolicyEvaluationCache implements PolicyEvaluationC
         } catch (IOException e) {
             throw new IllegalStateException("Failed to read policy evaluation result as JSON.", e);
         }
+        // Remove the Policy Evaluation Result out of values because we deserialized it
+        values.remove(values.size() - 1);
+
+        String zone = key.getZoneId();
+        // Add the resolved resource URI's stored in the decision to values only if the resource in the request
+        // does not match the resource stored in the cached decision so that we can check timestamps of all entities
+        // involved in the request.
+        Set<String> cachedResolvedResourceUris = cachedResult.getResolvedResourceUris();
+        if (!cachedResolvedResourceUris.equals(Collections.singleton(values.get(values.size() - 1)))) {
+            values.remove(values.size() - 1);
+            values.addAll(multiGet(cachedResolvedResourceUris.stream()
+                    .map(resourceUri -> resourceKey(zone, resourceUri)).collect(Collectors.toList())));
+        }
 
         if (isCachedRequestInvalid(values, new DateTime(cachedResult.getTimestamp()))) {
             delete(keys.get(keys.size() - 1));
@@ -77,11 +122,14 @@ public abstract class AbstractPolicyEvaluationCache implements PolicyEvaluationC
     }
 
     private List<String> assembleKeys(final PolicyEvaluationRequestCacheKey key) {
+        // Only need to assemble keys for policy sets, the subject, and the policy evaluation result.
+        // Resource-related information will be captured in the resolved resource URIs from the cached
+        // policy evaluation result.
         List<String> keys = new ArrayList<>();
         LinkedHashSet<String> policySetIds = key.getPolicySetIds();
         policySetIds.forEach(policySetId -> keys.add(policySetKey(key.getZoneId(), policySetId)));
-        keys.add(resourceKey(key.getZoneId(), key.getResourceId()));
         keys.add(subjectKey(key.getZoneId(), key.getSubjectId()));
+        keys.add(resourceKey(key.getZoneId(), key.getResourceId()));
         keys.add(key.toRedisKey());
         return keys;
     }
@@ -90,9 +138,9 @@ public abstract class AbstractPolicyEvaluationCache implements PolicyEvaluationC
             final List<String> keys, final List<String> values) {
         if (LOGGER.isDebugEnabled()) {
             LinkedHashSet<String> policySetIds = key.getPolicySetIds();
-            policySetIds.forEach(policySetId -> LOGGER.debug(String
-                    .format("Getting timestamp for policy set: '%s', key: '%s', timestamp:'%s'.", policySetId,
-                            keys.get(0), values.get(0))));
+            policySetIds.forEach(policySetId -> LOGGER
+                    .debug(String.format("Getting timestamp for policy set: '%s', key: '%s', timestamp:'%s'.",
+                            policySetId, keys.get(0), values.get(0))));
             LOGGER.debug("Getting timestamp for resource: '{}', key: '{}', timestamp:'{}'.", key.getResourceId(),
                     keys.get(1), values.get(1));
             LOGGER.debug("Getting timestamp for subject: '{}', key: '{}', timestamp:'{}'.", key.getSubjectId(),
@@ -101,36 +149,38 @@ public abstract class AbstractPolicyEvaluationCache implements PolicyEvaluationC
         }
     }
 
+    // Set's the policy evaluation key to the policy evaluation result in the cache
     @Override
     public void set(final PolicyEvaluationRequestCacheKey key, final PolicyEvaluationResult result) {
         try {
-            DateTime now = new DateTime();
-            result.setTimestamp(now.getMillis());
+            setEntityTimestamps(key, result);
+            result.setTimestamp(new DateTime().getMillis());
             String value = OBJECT_MAPPER.writeValueAsString(result);
             set(key.toRedisKey(), value);
-            setResourceTranslations(key.getZoneId(), result.getResolvedResourceUris(), key.getResourceId());
             LOGGER.debug("Setting policy evaluation to cache; key: '{}', value: '{}'.", key.toRedisKey(), value);
         } catch (IOException e) {
             throw new IllegalArgumentException("Failed to write policy evaluation result as JSON.", e);
         }
     }
 
-    @Override
-    public void setResourceTranslation(final String zoneId, final String fromResourceId, final String toResourceId) {
-        String fromKey = resourceTranslationKey(zoneId, fromResourceId);
-        String toKey = resourceKey(zoneId, toResourceId);
-        setResourceTranslation(fromKey, toKey);
-    }
+    private void setEntityTimestamps(final PolicyEvaluationRequestCacheKey key, final PolicyEvaluationResult result) {
+        // This ensures that if the timestamp for any entity involved in this decision is not in the cache at the time
+        // of this evaluation, it will be put there so that in subsequent evaluations, we will use the cached
+        // decision.
+        // We reset the timestamp to now for entities only if they do not exist in the cache so that we don't
+        // invalidate previous cached decisions.
+        String zoneId = key.getZoneId();
+        setSubjectIfNotExists(zoneId, key.getSubjectId());
 
-    @Override
-    public void setResourceTranslations(final String zoneId, final Set<String> fromResourceIds,
-            final String toResourceId) {
-        Set<String> fromKeys = new HashSet<>();
-        for (String fromResourceId : fromResourceIds) {
-            fromKeys.add(resourceTranslationKey(zoneId, fromResourceId));
+        Set<String> resolvedResourceUris = result.getResolvedResourceUris();
+        for (String resolvedResourceUri : resolvedResourceUris) {
+            setResourceIfNotExists(zoneId, resolvedResourceUri);
         }
-        String toKey = resourceKey(zoneId, toResourceId);
-        setResourceTranslations(fromKeys, toKey);
+
+        Set<String> policySetIds = key.getPolicySetIds();
+        for (String policySetId : policySetIds) {
+            setPolicySetIfNotExists(zoneId, policySetId);
+        }
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


FrankGasparovic:
Notes on new caching: 
https://devcloud.swcoe.ge.com/devspace/display/RMOCM/ACS+Decision+Caching

service/src/main/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationServiceImpl.java
```diff
@@ -150,7 +156,12 @@ public PolicyEvaluationResult evalPolicy(final PolicyEvaluationRequestV1 request
                             "Processed Policy Evaluation for: "
                                     + "resourceUri='%s', subjectIdentifier='%s', action='%s'," + " result='%s'",
                             uri, subjectIdentifier, action, result.getEffect()));
-            this.cache.set(key, result);
+            try {
+                this.cache.set(key, result);
+            } catch (Exception e) {
+                LOGGER.error(String.format(
+                        "Unable to set cache key '%s' to value '%s' due to redis exception", key, result), e);
+            }
```
irinaepshteyn: Error message should not be tied to redis.  Could have other cache implementations.

service/src/main/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationServiceImpl.java
```diff
@@ -118,7 +118,13 @@ public PolicyEvaluationResult evalPolicy(final PolicyEvaluationRequestV1 request
...
         }
 
-        PolicyEvaluationResult result = this.cache.get(key);
+        PolicyEvaluationResult result = null;
+        try {
+            result = this.cache.get(key);
+        } catch (Exception e) {
+            LOGGER.error(String.format("Unable to get cache key '%s'", key), e);
+        }
+
```
irinaepshteyn, line 124: Add a unit test that policy evaluation returns expected result if cache get or set fails

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
```diff
@@ -67,6 +88,11 @@ public PolicyEvaluationResult get(final PolicyEvaluationRequestCacheKey key) {
         } catch (IOException e) {
             throw new IllegalStateException("Failed to read policy evaluation result as JSON.", e);
         }
+        values.remove(values.size() - 1);
+
+        String zone = key.getZoneId();
+        values.addAll(multiGet(cachedResult.getResolvedResourceUris().stream()
+                .map(resourceUri -> resourceKey(zone, resourceUri)).collect(Collectors.toList())));
```
irinaepshteyn: Have you thought of including resolved URIs as part of evaluation result key to avoid another trip to the cache store?  Something like:
zoneId:resourceId:resolvedURIList:...
Are we limited on the length of the key?

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
```diff
@@ -79,10 +105,12 @@ public PolicyEvaluationResult get(final PolicyEvaluationRequestCacheKey key) {
     }
 
     private List<String> assembleKeys(final PolicyEvaluationRequestCacheKey key) {
+        // Only need to assemble keys for policy sets, the subject, and the policy evaluation result.
+        // Resource-related information will be captured in the resolved resource URIs from the cached
+        // policy evaluation result.
         List<String> keys = new ArrayList<>();
```
irinaepshteyn: If resolvedURI were part of evaluation result key we could assemble resource keys for resolved URIs here.

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
irinaepshteyn: Add a unit test that verifies that get returns null if one of the resolved resources change.  Resolved URI of the changed does not match resource URI in the request.

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
```diff
@@ -121,22 +148,21 @@ public void set(final PolicyEvaluationRequestCacheKey key, final PolicyEvaluatio
...
 
-    @Override
-    public void setResourceTranslation(final String zoneId, final String fromResourceId, final String toResourceId) {
-        String fromKey = resourceTranslationKey(zoneId, fromResourceId);
-        String toKey = resourceKey(zoneId, toResourceId);
-        setResourceTranslation(fromKey, toKey);
-    }
+    private void setEntityTimestamps(final PolicyEvaluationRequestCacheKey key, final PolicyEvaluationResult result) {
+        // This will lower the overall hit rate for policy evaluation but is required to ensure that if
+        // the cached timestamp for any entity involved in this decision is aged out, the cached decision will be used
```
irinaepshteyn: Did you mean WILL NOT be used?

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
```diff
@@ -67,6 +88,11 @@ public PolicyEvaluationResult get(final PolicyEvaluationRequestCacheKey key) {
         } catch (IOException e) {
             throw new IllegalStateException("Failed to read policy evaluation result as JSON.", e);
         }
+        values.remove(values.size() - 1);
```
irinaepshteyn: May be values.removeAll()?

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
```diff
@@ -107,11 +135,10 @@ private void logCacheGetDebugMessages(final PolicyEvaluationRequestCacheKey key,
     @Override
     public void set(final PolicyEvaluationRequestCacheKey key, final PolicyEvaluationResult result) {
         try {
-            DateTime now = new DateTime();
-            result.setTimestamp(now.getMillis());
+            setEntityTimestamps(key, result);
+            result.setTimestamp(new DateTime().getMillis());
             String value = OBJECT_MAPPER.writeValueAsString(result);
```
irinaepshteyn, line 220: Add a unit test that caches the eval result with resolved URIs and get returns the valid cache since timestamps are recorded.

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
irinaepshteyn: Add a unit test that verifies that get returns null if timestamps for any entity including resolved URIs are not present.

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
```diff
@@ -150,139 +176,136 @@ public void reset(final PolicyEvaluationRequestCacheKey key) {
...
+        String key = getKey.apply(zoneId, entityId);
+        setIfNotExists(key, timestampValue());
+    }
+
+    private void logSetEntityTimestampsDebugMessage(final String timestamp, final String key, final String entityId,
+            final EntityType entityType) {
         if (LOGGER.isDebugEnabled()) {
-            LOGGER.debug(String.format("Setting timestamp for poliy set '%s'; key: '%s', value: '%s' ", policySetId,
-                    key, timestamp));
+            LOGGER.debug("Setting timestamp for {} '{}'; key: '{}', value: '{}'", entityType, entityId, key, timestamp);
```
irinaepshteyn: Since you are using lazy string construction we could rely on debug to check the debug level.  No need for extra isDebugEnabled call.

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
```diff
@@ -150,139 +176,136 @@ public void reset(final PolicyEvaluationRequestCacheKey key) {
...
+        logSetEntityTimestampsDebugMessage(timestamp, key, entityId, entityType);
+        set(key, timestamp);
+    }
+
+    private void setEntityIfNotExists(final String zoneId, final String entityId,
+            final BiFunction<String, String, String> getKey) {
+        String key = getKey.apply(zoneId, entityId);
+        setIfNotExists(key, timestampValue());
+    }
+
```
irinaepshteyn: Is there a difference between setEntityIfNotExists and resetForEntity?
When eval result is being cached we could call resetForEntity to set the timestamp.  It will create a entity key if doesn't exist.

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
```diff
@@ -150,139 +176,136 @@ public void reset(final PolicyEvaluationRequestCacheKey key) {
...
-        multisetForResources(zoneId, resourceIds.stream().collect(Collectors.toList()));
+        multisetForResources(zoneId, new ArrayList<>(resourceIds));
     }
 
     @Override
     public void resetForResources(final String zoneId, final List<ResourceEntity> resourceEntities) {
-        multisetForResources(zoneId, resourceEntities.stream().map(resource -> resource.getResourceIdentifier())
+        multisetForResources(zoneId, resourceEntities.stream().map(ResourceEntity::getResourceIdentifier)
                 .collect(Collectors.toList()));
     }
```
irinaepshteyn: We are collecting resource Ids into a list and then build a map of keys inside multisetForResources.  Could we combine it all in one call? Map streamed resourceEntities to resourceKey and collect into a map with the timestamp logging debug message for each... :-)

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
```diff
@@ -150,139 +176,136 @@ public void reset(final PolicyEvaluationRequestCacheKey key) {
...
-        multisetForSubjects(zoneId, subjectIds.stream().collect(Collectors.toList()));
+        multisetForSubjects(zoneId, new ArrayList<>(subjectIds));
     }
 
     @Override
     public void resetForSubjects(final String zoneId, final List<SubjectEntity> subjectEntities) {
         multisetForSubjects(zoneId,
-                subjectEntities.stream().map(subject -> subject.getSubjectIdentifier()).collect(Collectors.toList()));
+                subjectEntities.stream().map(SubjectEntity::getSubjectIdentifier).collect(Collectors.toList()));
     }
```
irinaepshteyn: Same comment as for resetForResources above.  Could we combine resetForSubjects and multisetForSubjects in one long beautiful fluent java 8 call? :-)

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
```diff
@@ -150,139 +176,136 @@ public void reset(final PolicyEvaluationRequestCacheKey key) {
...
 
-    boolean havePrivilegeServiceAttributesChanged(final List<String> values,
-            final DateTime policyEvalTimestampUTC) {
-        for (int i = 0; i < values.size() - 1; i++) {
-            if (null == values.get(i)) {
-                continue;
+    boolean havePrivilegeServiceAttributesChanged(final List<String> values, final DateTime policyEvalTimestampUTC) {
+        for (String value : values) {
+            if (null == value) {
+                return true;
```
irinaepshteyn: Values were already tested for nulls by isRequestCached before calling this method.

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
```diff
@@ -150,139 +176,136 @@ public void reset(final PolicyEvaluationRequestCacheKey key) {
...
                 throw new IllegalStateException("Failed to read timestamp from JSON.", e);
             }
-
             if (invalidationTimestampUTC.isAfter(policyEvalTimestampUTC)) {
+                if (LOGGER.isDebugEnabled()) {
+                    LOGGER.debug(
+                            "Privilege service attributes have timestamp '%s' which is after "
+                                    + "policy evaluation timestamp '%s'",
+                            invalidationTimestampUTC, policyEvalTimestampUTC);
+                }
```
irinaepshteyn: Since you are using lazy string construction we could rely on debug to check the debug level. No need for extra isDebugEnabled call.  Will message formatting work with %s or should we use {}?

irinaepshteyn (commented):
Review is still in progress.  These are the initial comments...

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
irinaepshteyn: I take it back.  Resolved UIRs could not be part of the eval result key as they are not known at the time of the request.

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
FrankGasparovic: setEntityIfNotExists will set the timestamp for the object if the object is not in the cache. resetForEntity will reset the timestamp regardless if the value if the value is in the cache

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
FrankGasparovic: I do mean will, but I think the message is confusing. I rewrote it.

service/src/main/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationServiceImpl.java
FrankGasparovic: Fixed

service/src/main/java/com/ge/predix/acs/service/policy/evaluation/PolicyEvaluationServiceImpl.java
FrankGasparovic: Fixed

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
FrankGasparovic: Fixed

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
FrankGasparovic: testGetWithCacheHit covers this case

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
FrankGasparovic: Removed

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
```diff
@@ -67,6 +96,14 @@ public PolicyEvaluationResult get(final PolicyEvaluationRequestCacheKey key) {
...
             throw new IllegalStateException("Failed to read policy evaluation result as JSON.", e);
         }
+        // Remove the Policy Evaluation Result out of values because we deserialized it
+        values.remove(values.size() - 1);
+
+        String zone = key.getZoneId();
+        // Add the resolved resource URI's stored in the cachedResult to values so that we can check timestamps of
+        // all entities involved in the request.
+        values.addAll(multiGet(cachedResult.getResolvedResourceUris().stream()
+                .map(resourceUri -> resourceKey(zone, resourceUri)).collect(Collectors.toList())));
```
irinaepshteyn: One optimization we could do here is to compare resource from request with resolved resource.  If policy doesn't contain any template URIs then request resource and resolved resource are the same.  In that case we don't have to perform a multiGet again.  In that case when assembling keys for the first call to multiGet we need to include a resource from request again.

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/PolicyEvaluationCache.java
```diff
@@ -29,24 +29,26 @@
...
-
-    void setResourceTranslations(String zoneId, Set<String> fromResourceIds, String toResourceId);
-
     void reset();
 
     void reset(PolicyEvaluationRequestCacheKey key);
 
     void resetForPolicySet(String zoneId, String policySetId);
 
+    void setPolicySetIfNotExists(String zoneId, String policySetId);
```
irinaepshteyn: setPolicySetIfNotExists, setResourceIfNotExists, setSubjectIfNotExists do not need to be part of cache interface.  These methods are implemented and used by AbstractPolicyEvaluationCache only.  However, setIfNotExists should be defined by interface and implemented by each cache implementation.

irinaepshteyn (commented):
More comments...

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/PolicyEvaluationCache.java
FrankGasparovic: I agree that setPolicySetIfNotExists, setResourceIfNotExists and setSubjectIfNotExists should not be part of the cache interface (because these are only used internally in the AbstractPolicyEvaluationCache). But why should setIfNotExists be apart of the interface if it is also only use internally?

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/RedisPolicyEvaluationCache.java
```diff
@@ -123,21 +97,7 @@ public void setCachedEvalTimeToLiveSeconds(final long cachedEvalTimeToLiveSecond
...
-            public List<Object> doInRedis(final RedisConnection connection) throws DataAccessException {
-                StringRedisConnection stringRedisConn = new DefaultStringRedisConnection(connection);
-                for (String fromKey : fromKeys) {
-                    stringRedisConn.sAdd(fromKey, toKey);
-                }
-                return null;
-            }
-        });
+    void setIfNotExists(final String key, final String value) {
+        this.redisTemplate.boundValueOps(key).setIfAbsent(value);
```
irinaepshteyn, line 100: Should we set an expiration timeout on the key similar to what's done in line #89?  The same comment applies to set function above.  Timeout is only set on the key for eval result and not for any other.

FrankGasparovic: Timeout should not be set for anything other then Policy Evaluation. This is because we would then be unnecessarily reevaluating requests when entities have no changed.

service/src/test/java/com/ge/predix/acs/policy/evaluation/cache/InMemoryPolicyEvaluationCacheTest.java
```diff
@@ -73,9 +70,7 @@ public void testSetPolicyPolicySetChangedTimestamp() throws Exception {
         this.cache.set(key, value);
 
         Map<String, String> evalCache = (Map<String, String>) getInternalState(this.cache, "evalCache");
-        assertEquals(evalCache.size(), 0);
-        Map<String, String> timestampCache = (Map<String, String>) getInternalState(this.cache, "timestampCache");
-        assertEquals(timestampCache.size(), 1);
+        assertEquals(evalCache.size(), 1);
```
irinaepshteyn, line 73: May be rename the test to testSetPolicySetChangedTimestamp?

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
FrankGasparovic: Added

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
FrankGasparovic: Fixed

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
FrankGasparovic: Fixed

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
FrankGasparovic: Fixed

service/src/test/java/com/ge/predix/acs/policy/evaluation/cache/InMemoryPolicyEvaluationCacheTest.java
irinaepshteyn: Current implementation of the in-memory cache stores eval result and timestamps in the same map, not in different maps as before.  There is no need to have different tests for setting eval result and setting timestamps.

service/src/test/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCacheTest.java
```diff
@@ -88,9 +91,13 @@ public void testGetWithCacheMiss() {
     public void testGetWithCacheHit() {
```
irinaepshteyn, line 108: Add a test to assert that get returns cached result if eval result is cached but not timestamps for entities.

service/src/test/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCacheTest.java
```diff
@@ -119,14 +129,18 @@ public void testGetWithPolicyInvalidation() throws Exception {
...
 
     @Test
     public void testGetWithResetForMultiplePolicySets() throws Exception {
 
         PolicyEvaluationRequestV1 request = new PolicyEvaluationRequestV1();
+        this.cache.resetForSubject(ZONE_NAME, AGENT_MULDER);
+        this.cache.resetForResource(ZONE_NAME, XFILES_ID);
+        this.cache.resetForPolicySet(ZONE_NAME, POLICY_ONE.getName());
+        this.cache.resetForPolicySet(ZONE_NAME, POLICY_TWO.getName());
         request.setAction(ACTION_GET);
```
irinaepshteyn, line 161: Do we really need to call resets for each entity i every test?  Get will return cached result if no timestamps.

irinaepshteyn (changes requested):
Done reviewing.

service/src/test/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCacheTest.java
FrankGasparovic: We do, get will return null if no timestamps.

service/src/test/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCacheTest.java
FrankGasparovic: If eval is cached but timestamps for entities are not, then get should return null

service/src/test/java/com/ge/predix/acs/policy/evaluation/cache/InMemoryPolicyEvaluationCacheTest.java
FrankGasparovic: I do feel like there is value in this test because of the logic to check the key format.

FrankGasparovic:
Siva ran performance tests and he saw a 3ms increase in Policy Evaluation calls

sanjeevchopra:
@anubhavi25 @FrankGasparovic - pls review my 2nd commit

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
```diff
@@ -80,59 +79,122 @@ public String toString() {
...
+        PolicyEvaluationResult cachedEvalResult = toPolicyEvaluationResult(cachedEvalResultString);
+        
+        List<String> invalidationTimeStamps = new ArrayList<>();
+        invalidationTimeStamps.add(cachedEntries.getSubjectLastModified());
+        invalidationTimeStamps.addAll(cachedEntries.getPolicySetsLastModified());
+
+        Set<String> cachedResolvedResourceUris = cachedEvalResult.getResolvedResourceUris();
+        //is requested resource id same as resolved resource uri ?
+        if (cachedResolvedResourceUris.size() == 1
+                && cachedResolvedResourceUris.toArray()[0].equals(key.getResourceId())) {
```
FrankGasparovic: Use cachedResolvedResourceUris.iterator().next() vs toArray() so that we don't have to copy contents to array.

FrankGasparovic (commented):
Looks good.

service/src/main/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCache.java
```diff
@@ -80,59 +79,122 @@ public String toString() {
...
-    }
+        String getDecisionKey() {
+            return decisionKey;
+        }
+
+        String getSubjectLastModified() {
+            return entryValues.get(lastValueIndex - 2);
+        }
+
+        String getRequestedResourceLastModified() {
```
anubhavi25, line 189: Changing the name to just `getResourceLastModified` should suffice (and is also consistent with the naming convention used for the other `get*LastModified` methods).

anubhavi25 (commented):
Looks good for the most part 👍

sanjeevchopra:
Performance tests show no regression. A second test was run which only writes resource and subject keys (which have no TTL) to confirm that eviction is working as expected.

PR #118, opened 2017-04-11
US107561: Enforce max storage per adapter attribute cache entry

FrankGasparovic:


model/src/main/java/com/ge/predix/acs/model/Attribute.java
```diff
@@ -114,4 +118,19 @@ public boolean equals(final Object obj) {
     public String toString() {
         return "Attribute [issuer=" + this.issuer + ", name=" + this.name + ", value=" + this.value + "]";
     }
+
+    public int size() {
```
anubhavi25: It might be cleaner to always recompute the size using a local variable in this method instead of having side-effects in the setters above (where you do `this.size = 0;`), esp. since there seems to be no performance improvements in the way that you have it now. This would also allow getting rid of the private member variable `size`.

model/src/main/java/com/ge/predix/acs/rest/attribute/adapter/AttributesResponse.java
```diff
@@ -33,4 +39,28 @@ public String getId() {
     public void setId(final String id) {
         this.id = id;
     }
+
+    public int checkSizeLimit(final int maxNumberOfAttributes, final int maxSizeOfAttributesInBytes) {
```
anubhavi25: See comment above re. recomputing using local variables and removing private member variables.

Also maybe the following signature is more clear since what we really care about is whether or not we've exceeded set thresholds:

`public boolean sizeLimitsExceeded ...`

This (and related refactoring inside this method) would also allow us to get rid of the reliance on `Integer.MAX_VALUE`

model/src/main/java/com/ge/predix/acs/rest/attribute/adapter/AttributesResponse.java
```diff
@@ -33,4 +39,28 @@ public String getId() {
...
+
+        if (this.cachedMaxNumberOfAttributes == maxNumberOfAttributes
```
anubhavi25: This condition can go away once limits are always recomputed

model/src/test/java/com/ge/predix/acs/rest/AttributeResponseTest.java
```diff
@@ -0,0 +1,43 @@
...
+        AttributesResponse attributesResponse = createAttributesResponse(1);
+        Assert.assertEquals(attributesResponse.checkSizeLimit(100, 1), Integer.MAX_VALUE);
+
+    }
+
+    private AttributesResponse createAttributesResponse(final int numberOfAttributes) {
+        AttributesResponse attributesResponse = new AttributesResponse();
+        Set<Attribute> attributeSet = new HashSet<>();
+        for (int i = 0; i < numberOfAttributes; i++) {
+            attributeSet.add(new Attribute("issuer" + Integer.toString(i), "name" + Integer.toString(i),
```
anubhavi25: Assign `Integer.toString(i)` to a local variable to avoid executing it multiple times?

service/src/main/java/com/ge/predix/acs/attribute/readers/AttributeRetrievalException.java
```diff
@@ -8,7 +8,15 @@ public AttributeRetrievalException(final String message, final Throwable cause)
...
+    public AttributeRetrievalException(final String message) {
+        super(message);
+    }
+
     static String getAdapterErrorMessage(final String adapterEndpoint) {
         return MessageFormat.format("Couldn''t get attributes from the adapter with endpoint ''{0}''", adapterEndpoint);
     }
+
+    static String getStorageErrorMessage(final String id) {
+        return String.format("Total size of attributes or number of attributes to large for id: '%s'", id);
```
anubhavi25: Typo: to -> too.

Also is it worth mentioning the thresholds in the message at the time this exception is thrown to clue the user in (in case something can be done to proactively limit the attributes on the adapter side)?

service/src/main/java/com/ge/predix/acs/attribute/readers/ExternalAttributeReader.java
```diff
@@ -21,33 +21,45 @@
 import com.ge.predix.acs.rest.attribute.adapter.AttributesResponse;
 
 public abstract class ExternalAttributeReader implements AttributeReader {
+
+    @Value("${MAX_NUMBER_OF_ATTRIBUTES:1500}")
+    private int maxNumberOfAttributes;
+
+    @Value("${MAX_SIZE_OF_ATTRIBUTES:500000}")
```
anubhavi25: Use `MAX_SIZE_OF_ATTRIBUTES_IN_BYTES` (similar to the name of the member variable it annotates)

service/src/main/java/com/ge/predix/acs/attribute/readers/ExternalAttributeReader.java
```diff
@@ -92,20 +105,36 @@ private OAuth2RestTemplate getAdapterOauth2RestTemplate(final AttributeAdapterCo
...
+            if (attributesResponse.checkSizeLimit(maxNumberOfAttributes, maxSizeOfAttributesInBytes)
+                    == Integer.MAX_VALUE) {
+                if (LOGGER.isDebugEnabled()) {
+                    LOGGER.debug(AttributeRetrievalException.getStorageErrorMessage(identifier));
+                }
+                attributesWrapper.setException(new AttributeRetrievalException(
+                        AttributeRetrievalException.getStorageErrorMessage(identifier)));
+                return attributesWrapper;
             }
+            attributesWrapper.setAttributes(attributesResponse.getAttributes());
```
anubhavi25: Combine this line with line 111 since that's the only place it seems like we need to set attributes:

`attributesWrapper.setAttributes(adapter.getForEntity(adapterUrl, AttributesResponse.class).getBody().getAttributes());`

If you do this, you don't need the `AttributesResponse` local variable.

model/src/test/java/com/ge/predix/acs/rest/AttributeResponseTest.java
```diff
@@ -0,0 +1,43 @@
...
+                    "value" + Integer.toString(i)));
+        }
+        attributesResponse.setAttributes(attributeSet);
+        return attributesResponse;
```
anubhavi25: `return new AttributesResponse().setAttributes(attributeSet);` should get rid of the local variable?

model/src/test/java/com/ge/predix/acs/rest/AttributeResponseTest.java
```diff
@@ -0,0 +1,43 @@
...
+public class AttributeResponseTest {
+
+    @Test
+    public void testCheckSizeLimit() {
+        AttributesResponse attributesResponse = createAttributesResponse(1);
+        Assert.assertEquals(attributesResponse.checkSizeLimit(100, 100), 17);
+    }
+
+    @Test
+    public void testCheckSizeLimitMaxAttributes() {
```
anubhavi25: Rename to `testMaxNumberOfAttributesExceeded` for clarity?

model/src/test/java/com/ge/predix/acs/rest/AttributeResponseTest.java
```diff
@@ -0,0 +1,43 @@
...
+        AttributesResponse attributesResponse = createAttributesResponse(2);
+        Assert.assertEquals(attributesResponse.checkSizeLimit(1, 100), Integer.MAX_VALUE);
+    }
+
+    @Test
+    public void testCheckSizeLimitMaxAttributesBytes() {
```
anubhavi25: Rename to `testMaxSizeOfAttributesExceeded` for clarity?

model/src/test/java/com/ge/predix/acs/rest/AttributeResponseTest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import com.ge.predix.acs.model.Attribute;
+import com.ge.predix.acs.rest.attribute.adapter.AttributesResponse;
+
+public class AttributeResponseTest {
+
+    @Test
+    public void testCheckSizeLimit() {
```
anubhavi25: Rename to `testAttributesWithinSizeLimits` for clarity?

model/src/test/java/com/ge/predix/acs/model/AttributeTest.java
```diff
@@ -0,0 +1,27 @@
+package com.ge.predix.acs.model;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class AttributeTest {
```
anubhavi25: As discussed, it's debatable how much value POJO-related unit tests provide. Maybe we should create such tests once we've identified gaps via code coverage?

model/src/main/java/com/ge/predix/acs/rest/attribute/adapter/AttributesResponse.java
```diff
@@ -33,4 +39,28 @@ public String getId() {
...
+        } else {
+            for (Attribute attribute : this.attributes) {
+                this.cachedSize += attribute.size();
+                if (this.cachedSize > maxSizeOfAttributesInBytes) {
+                    this.cachedSize = Integer.MAX_VALUE;
+                    break;
+                }
+            }
+        }
+        this.cachedMaxNumberOfAttributes = maxNumberOfAttributes;
```
anubhavi25: These assignments can go away once limits are always recomputed

anubhavi25 (changes requested):

model/src/main/java/com/ge/predix/acs/model/Attribute.java
FrankGasparovic: Fixed

sanjeevchopra: Because this class is visible to ACS users, I would prefer not to have this here because it is an internal concern.  Also, the size restrictions only apply to attributes which are sourced from external adapters.
* We should not modify acs-model for this change.

model/src/main/java/com/ge/predix/acs/rest/attribute/adapter/AttributesResponse.java
sanjeevchopra: same concern about not exposing this to ACS users. Especially because these limits only apply for attributes sourced from external sources.
We should not change anything in acs-model for this.

service/src/main/java/com/ge/predix/acs/attribute/readers/AttributesWrapper.java
```diff
@@ -0,0 +1,35 @@
+package com.ge.predix.acs.attribute.readers;
+
+import java.util.Set;
+
+import com.ge.predix.acs.model.Attribute;
+
+public class AttributesWrapper {
```
sanjeevchopra: Wrapper for what purpose ?  How about  CachedAttributes

service/src/main/java/com/ge/predix/acs/attribute/readers/AttributesWrapper.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public AttributesWrapper(final Set<Attribute> attributes) {
+        this.attributes = attributes;
+    }
+
+    public AttributesWrapper(final RuntimeException exception) {
+        this.exception = exception;
+    }
+
+    private RuntimeException exception;
```
sanjeevchopra: * Consider using a `state` field which would return a set of well known states, instead of exception.
* minor -  pls move the field declarations to the top

service/src/main/java/com/ge/predix/acs/attribute/readers/ExternalAttributeReader.java
```diff
@@ -92,20 +103,36 @@ private OAuth2RestTemplate getAdapterOauth2RestTemplate(final AttributeAdapterCo
...
-                        AttributesResponse.class);
-                attributes = attributesResponse.getBody().getAttributes();
+                attributesResponse = adapter.getForEntity(adapterUrl, AttributesResponse.class).getBody();
             } catch (final Throwable e) {
-                throw new AttributeRetrievalException(AttributeRetrievalException.getAdapterErrorMessage(adapterUrl),
-                        e);
+                if (LOGGER.isDebugEnabled()) {
+                    LOGGER.debug(AttributeRetrievalException.getAdapterErrorMessage(adapterUrl), e);
+                }
+                attributesWrapper.setException(
```
sanjeevchopra: don't think ACS should treat this  the same as a response size being too large.  This needs to be handled with a circuit breaker (separate story). For now, it should just retry on the next request ?

sanjeevchopra (changes requested):
Is there a test case, when the response is too large ? I don't see it.

FrankGasparovic:
Build pass https://predix1.jenkins.build.ge.com/job/Predix-Security/job/ACS/job/ACS-PullRequest/job/acs-integration-pullrequest/badge/icon

PR #120, opened 2017-04-24
US67524: Added health checks for dependent services

anubhavi25:


anubhavi25:
NOTE: This PR has been tested with `public*` and `predix-local*` Maven profiles but not `predix-cloud*`

service/src/main/resources/security-config.xml
```diff
@@ -219,7 +219,20 @@
...
     </http>
 
+    <!-- Authorization Configuration For Monitoring APIs -->
+    <http pattern="/health*" request-matcher="ant" create-session="stateless"
+          xmlns="http://www.springframework.org/schema/security"
+          disable-url-rewriting="true" use-expressions="true"
+          entry-point-ref="preAuthenticationEntryPoint">
+        <csrf disabled="true"/>
+
+        <intercept-url pattern="/health*" access="isAnonymous() or hasAnyAuthority('acs.monitoring')"/>
```
irinaepshteyn, line 229: We should allow either anonymous access for a limited health status or fully authenticated access with monitoring authority for a complete health status:
isAnonymous() or (isFullyAuthenticated() and hasAnyAuthority('acs.monitoring'))

service/src/main/resources/security-config.xml
```diff
@@ -219,7 +219,20 @@
...
+        <custom-filter ref="oauth2remoteTokenFilter" position="PRE_AUTH_FILTER"/>
+        <custom-filter ref="managementSecurityRoleFilter" after="BASIC_AUTH_FILTER"/>
```
irinaepshteyn, line 231: I am not clear why we need additional managementSecurityRoleFilter.  Token will be validated by fast token service invoked by oauth2remoteTokenFilter.  Required authorities will be verified by spring security hasAnyAuthority.  Please clarify

acs-integration-tests/src/test/java/com/ge/predix/integration/test/AcsMonitoringIT.java
```diff
@@ -0,0 +1,96 @@
...
+        ResponseEntity<String> response = this.hitHealthCheckUrl(this.trustedUaaTokenUrl, this.authorizedClientId);
+        Assert.assertEquals(response.getStatusCode(), HttpStatus.OK);
+        JsonNode responseJson = OBJECT_MAPPER.readTree(response.getBody());
+        Assert.assertTrue(responseJson.size() > 1);
+        Assert.assertTrue(responseJson.hasNonNull(AcsMonitoringUtilities.STATUS));
+        List<String> activeProfiles = Arrays.asList(this.environment.getActiveProfiles());
+        Assert.assertTrue(activeProfiles.contains("titan") == responseJson.hasNonNull("graphDb"));
+        Assert.assertTrue(activeProfiles.contains("predix") == responseJson.hasNonNull("zac"));
+        Assert.assertTrue((activeProfiles.contains("redis") || activeProfiles.contains("cloud-redis")) == responseJson
+                .hasNonNull("decisionCache"));
```
irinaepshteyn: Add assertion that response contains UAA and Postgres DB statuses

service/src/main/java/com/ge/predix/acs/monitoring/AcsHealthAggregator.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+@Component
+// Modified version of org.springframework.boot.actuate.health.OrderedHealthAggregator
+public class AcsHealthAggregator implements HealthAggregator {
+
```
irinaepshteyn, line 19: Since it's a modified version, should AcsHealthAggregator extend OrderedHealthAggregator?

irinaepshteyn: Why can't we just use OrderedHealthAggregator?

irinaepshteyn (commented):
Review is still in progress.  These are my initial question/comments.

service/src/main/java/com/ge/predix/acs/monitoring/UaaHealthIndicator.java
```diff
@@ -12,57 +12,60 @@
...
+    private AcsMonitoringUtilities.HealthCode check() {
+        AcsMonitoringUtilities.HealthCode healthCode;
 
-            LOGGER.debug("Checking UAA Status");
+        try {
+            LOGGER.debug("Checking UAA status using URL: {}", this.uaaCheckHealthUrl);
             this.uaaTemplate.getForObject(this.uaaCheckHealthUrl, String.class);
-            return SUCCESS_CHECK;
-
+            healthCode = AcsMonitoringUtilities.HealthCode.AVAILABLE;
```
irinaepshteyn, line 56: Should we check that response is 'ok'?  What health code should be returned is response is not 'ok' and no exception is thrown?

service/src/test/java/com/ge/predix/acs/monitoring/UaaHealthIndicatorTest.java
```diff
@@ -29,34 +29,41 @@
...
-        return new Object[][] { new Object[] { mockRestWithUp(), Status.UP },
-                new Object[] { mockRestWithException(), new Status(AcsMonitoringConstants.UAA_OUT_OF_SERVICE) }, };
+    public Object[][] statuses() {
+        return new Object[][] { new Object[] { mockRestWithUp(), Status.UP, null },
+
+                { mockRestWithException(new RestClientException("")), Status.DOWN,
+                        AcsMonitoringUtilities.HealthCode.UNREACHABLE },
+
+                { mockRestWithException(new RuntimeException()), Status.DOWN,
+                        AcsMonitoringUtilities.HealthCode.ERROR }, };
```
irinaepshteyn, line 56: Add a test for the use case when UAA health returns something other than 'ok' and no exception is thrown.

service/src/main/java/com/ge/predix/acs/monitoring/UaaHealthIndicator.java
irinaepshteyn: Assign default health code "ERROR"

service/src/test/java/com/ge/predix/acs/monitoring/ZacHealthIndicatorTest.java
```diff
@@ -0,0 +1,81 @@
...
+                        AcsMonitoringUtilities.HealthCode.INVALID_JSON },
+
+                { mockRestWithInvalidJson("{\"status\":\"BOGUS\"}"), Status.DOWN,
+                        AcsMonitoringUtilities.HealthCode.ERROR },
+
+                { mockRestWithException(new RestClientException("")), Status.DOWN,
+                        AcsMonitoringUtilities.HealthCode.UNREACHABLE },
+
+                { mockRestWithException(new RuntimeException()), Status.DOWN,
+                        AcsMonitoringUtilities.HealthCode.ERROR }, };
```
irinaepshteyn, line 62: Add a use case when returned status is DOWN and no exception is thrown.

service/src/test/java/com/ge/predix/acs/monitoring/ZacHealthIndicatorTest.java
```diff
@@ -0,0 +1,81 @@
...
+        } else {
+            Assert.assertEquals(healthCode,
+                    zacHealthIndicator.health().getDetails().get(AcsMonitoringUtilities.CODE_KEY));
+        }
+    }
+
+    @DataProvider
+    public Object[][] statuses() {
+        return new Object[][] { new Object[] { mockRestWithUp(), Status.UP, null },
+
```
irinaepshteyn: Based on implementation, when ZAC returned health status is UP, health code is set to AVAILABLE, not null.  Please check this test.

service/src/main/java/com/ge/predix/acs/monitoring/DecisionCacheHealthIndicator.java
```diff
@@ -0,0 +1,59 @@
...
+
+    RedisConnection getRedisConnection() {
+        return RedisConnectionUtils.getConnection(this.redisConnectionFactory);
+    }
+
+    private AcsMonitoringUtilities.HealthCode check() {
+        AcsMonitoringUtilities.HealthCode healthCode;
+
+        RedisConnection connection = getRedisConnection();
+
```
irinaepshteyn: Move getting Redis connection inside the try/catch block in case some exceptions happen.

service/src/main/java/com/ge/predix/acs/monitoring/DecisionCacheHealthIndicator.java
```diff
@@ -0,0 +1,59 @@
...
+        return AcsMonitoringUtilities.health(this::check, DESCRIPTION);
+    }
+
+    RedisConnection getRedisConnection() {
+        return RedisConnectionUtils.getConnection(this.redisConnectionFactory);
+    }
+
+    private AcsMonitoringUtilities.HealthCode check() {
+        AcsMonitoringUtilities.HealthCode healthCode;
+
```
irinaepshteyn, line 57: Set health code default

service/src/test/java/com/ge/predix/acs/monitoring/DecisionCacheHealthIndicatorTest.java
```diff
@@ -0,0 +1,63 @@
...
+        } else {
+            Assert.assertEquals(healthCode,
+                    decisionCacheHealthIndicator.health().getDetails().get(AcsMonitoringUtilities.CODE_KEY));
+        }
+    }
+
+    @DataProvider
+    public Object[][] statuses() {
+        return new Object[][] { new Object[] { mockCacheWithUp(), Status.UP, null },
+
```
irinaepshteyn: Health code should be AVAILABLE when cache is UP.  Please check this test case.

service/src/main/java/com/ge/predix/acs/monitoring/DecisionCacheHealthIndicator.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.context.annotation.Profile;
+import org.springframework.data.redis.connection.RedisConnection;
+import org.springframework.data.redis.connection.RedisConnectionFactory;
+import org.springframework.data.redis.core.RedisConnectionUtils;
+import org.springframework.stereotype.Component;
+import org.springframework.util.Assert;
+
+@Component
+@Profile({ "cloud-redis", "redis" })
+@ConditionalOnEnabledHealthIndicator("decision.cache")
```
irinaepshteyn: We should not perform a health check but still report a status indicating that cache is disabled or redis health check is disabled.  Such state is defined by the following environment variables:
ENABLE_CACHING
ENABLED_REDIS_HEALTH_CHECK

service/src/main/java/com/ge/predix/acs/monitoring/GraphDbHealthIndicator.java
```diff
@@ -0,0 +1,102 @@
...
+import org.springframework.stereotype.Component;
+
+import java.util.Collections;
+import java.util.Set;
+import java.util.UUID;
+
+@Component
+@Profile({ "titan" })
+public class GraphDbHealthIndicator implements HealthIndicator {
+
```
irinaepshteyn, line 22: Should we report a special status depending on TITAN_ENABLE_CASSANDRA setting, either cassandra or in-memory database is used?
[GraphConfig.java#L86-L102](https://github.com/predix/acs/blob/develop/service/src/main/java/com/ge/predix/acs/config/GraphConfig.java#L86-L102)

service/src/main/java/com/ge/predix/acs/monitoring/GraphDbHealthIndicator.java
```diff
@@ -0,0 +1,102 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(GraphDbHealthIndicator.class);
+    private static final String ERROR_MESSAGE_FORMAT = "Unexpected exception while checking graph database status: {}";
+    static final String DESCRIPTION = "Health check performed by attempting to create a zone entity, a parent resource,"
+            + " a child resource, and getting the descendants of the parent resource (including the parent itself)";
+
```
irinaepshteyn: Seems like we are running a full fledged functional test for hierarchical attributes here. Would it be simpler to just autowire graph and query for a version node? 
[GraphConfig.java#L111](https://github.com/predix/acs/blob/develop/service/src/main/java/com/ge/predix/acs/config/GraphConfig.java#L111)

service/src/test/java/com/ge/predix/acs/monitoring/AcsHealthAggregatorTest.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Test(dataProvider = "statuses")
+    public void defaultOrderWithDegradedStatus(final Status expectedStatus, final List<Status> actualStatuses) {
+        Map<String, Health> healths = new HashMap<>();
+        healths.put("h1", new Health.Builder().status(actualStatuses.get(0)).build());
+        healths.put("h2", new Health.Builder().status(actualStatuses.get(1)).build());
+        healths.put("h3", new Health.Builder().status(actualStatuses.get(2)).build());
+        Assert.assertEquals(this.healthAggregator.aggregate(healths).getStatus(), expectedStatus);
+    }
```
irinaepshteyn, line 64: Add a test case for:
healthAggregator.aggregate(empty healths); expected status is UKNOWN
healthAggregator.aggregate(healths contains one entry with 'cache' and status DOWN); expected status DEGRADED_STATUS
healthAggregator.aggregate(healths contains one entry with CUSTOM status); expected status UNKNOWN

irinaepshteyn (changes requested):
Done reviewing.  Please address my comments.

acs-integration-tests/src/test/java/com/ge/predix/integration/test/AcsMonitoringIT.java
anubhavi25: Added

service/src/main/java/com/ge/predix/acs/monitoring/AcsHealthAggregator.java
anubhavi25: I originally intended to use `OrderedHealthAggregator` but unfortunately the `AbstractHealthAggregator#aggregate` method that requires customization in this case has been marked `final` by the implementors and thus a) can't be overridden and b) isn't customizable in the way we want using the other `protected` methods provided in that class. Moreover, my static analyzer flagged certain legitimate problems in `OrderedHealthAggregator` that I decided to fix in a modified version of the class.

service/src/main/java/com/ge/predix/acs/monitoring/DecisionCacheHealthIndicator.java
anubhavi25: Agreed added new statuses for both and associated unit tests. The overall status will be `DEGRADED` if either caching or the Redis health check is disabled via environment variables.

service/src/main/java/com/ge/predix/acs/monitoring/DecisionCacheHealthIndicator.java
anubhavi25: No need to set a default value because under no circumstance would `healthCode` *not* be set

service/src/main/java/com/ge/predix/acs/monitoring/DecisionCacheHealthIndicator.java
anubhavi25: Moved

service/src/main/java/com/ge/predix/acs/monitoring/GraphDbHealthIndicator.java
anubhavi25: I thought about this too and under normal production circumstances we only care to check the health of the underlying graph storage engine (Cassandra in this case).

service/src/main/java/com/ge/predix/acs/monitoring/GraphDbHealthIndicator.java
anubhavi25: Yea let's go with that 👍

service/src/main/java/com/ge/predix/acs/monitoring/UaaHealthIndicator.java
anubhavi25: The health code is *always* "ok" so checking the response doesn't seem to provide much value: https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/health/HealthzEndpoint.java. Also no need to set a default value because under no circumstance would `healthCode` *not* be set and if a non-expected exception is thrown, it will be set to `ERROR` anyways.

service/src/main/resources/security-config.xml
anubhavi25: If we end up providing a health check dashboard in the future, using `isFullyAuthenticated()` would be too strict since it wouldn't allow for cases where the user has already logged in (checked via Spring Security's `isAuthenticated()` or `isRememberMe()` methods).

service/src/main/resources/security-config.xml
anubhavi25: See https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-monitoring.html#production-ready-sensitive-endpoints. As mentioned in the docs, Spring Boot Actuator requires the general `HttpServletRequest#isUserInRole` method to be overridden by at least one filter in the filter chain to determine whether or not a user can access sensitive endpoints. While it might overlap functionally with Spring Security's `hasAnyAuthority` method, `ManagementSecurityRoleFilter#doFilterInternal` does exactly that by telling Spring Boot Actuator to check that `acs.monitoring` is one of the authorities in the user's JWT token. `FastTokenService` doesn't currently use `isUserInRole` anywhere in its filter chain.

service/src/test/java/com/ge/predix/acs/monitoring/AcsHealthAggregatorTest.java
anubhavi25: Added

service/src/test/java/com/ge/predix/acs/monitoring/DecisionCacheHealthIndicatorTest.java
anubhavi25: Fixed

service/src/test/java/com/ge/predix/acs/monitoring/UaaHealthIndicatorTest.java
anubhavi25: See other comment re. UAA never returning a status other than "ok".

service/src/test/java/com/ge/predix/acs/monitoring/ZacHealthIndicatorTest.java
anubhavi25: Fixed

service/src/test/java/com/ge/predix/acs/monitoring/ZacHealthIndicatorTest.java
anubhavi25: This can only happen when a non-`UP` `status` is returned in the ZAC response and is already covered above by the test with `mockRestWithInvalidJson("{\"status\":\"BOGUS\"}")`.

anubhavi25:
@irinaepshteyn please do a final review since there are changes in addition to what you requested.

anubhavi25:
[Builds #573-576 Passing (`predix-cloud*` test failures are not related to this PR)](https://predix1.jenkins.build.ge.com/job/Predix-Security/job/ACS/job/ACS-PullRequest/job/acs-integration-pullrequest/)

service/src/main/java/com/ge/predix/acs/monitoring/DecisionCacheHealthIndicator.java
irinaepshteyn: Setting overall status to DEGRADED might trigger an alarm.  May be it's better to set overall status to UP but redis status DISABLED in case we need to temporarily disable caching but don't want any alarms.

service/src/main/java/com/ge/predix/acs/monitoring/GraphDbHealthIndicator.java
irinaepshteyn: May be then if cassandra is disabled we should return graph db status CASSANDRA_DISABLED or INMEMORY?

anubhavi25: Sure changed to using an `IN_MEMORY` status if `TITAN_ENABLE_CASSANDRA` is set to `false`.

service/src/main/java/com/ge/predix/acs/monitoring/DecisionCacheHealthIndicator.java
anubhavi25: Yea that's probably better.. changed to using `UP` instead as the overall status if either caching or the Redis health check is disabled via environment variables.

irinaepshteyn (approved):

PR #132, opened 2017-06-30
US140943: Code changes based on review of Caching for connectors

FrankGasparovic:


anurag-gujral (approved):
Looks good for most part thanks for making the changes. May be later we can remove duplicacy between CloudRedisProperties and LocalRedisProperties.

PR #133, opened 2017-06-30
US140969: Modify ACS graph repositories to use TinkerPop3 GraphTraversalSource instead of Graph object

FrankGasparovic:
Signed-off-by: Chris <chris.ruppelt@ge.com>

### DIFF
--- a/service/src/main/java/com/ge/predix/acs/privilege/management/dao/GraphGenericRepository.java
+++ b/service/src/main/java/com/ge/predix/acs/privilege/management/dao/GraphGenericRepository.java
@@ -15,9 +15,9 @@ import java.util.stream.Collectors;
 
 import org.apache.commons.lang.NotImplementedException;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
 import org.apache.tinkerpop.gremlin.structure.Direction;
 import org.apache.tinkerpop.gremlin.structure.Edge;
-import org.apache.tinkerpop.gremlin.structure.Graph;
 import org.apache.tinkerpop.gremlin.structure.T;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.apache.tinkerpop.gremlin.structure.VertexProperty;
@@ -51,7 +51,7 @@ public abstract class GraphGenericRepository<E extends ZonableEntity> implements
     public static final String VERSION_VERTEX_LABEL = "version";
 
     @Autowired
-    private Graph graph;
+    private GraphTraversalSource graphTraversal;
 
     @Value("${TITAN_TRAVERSAL_LIMIT:256}")
     private long traversalLimit = 256;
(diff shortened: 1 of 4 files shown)


service/src/main/java/com/ge/predix/acs/privilege/management/dao/GraphGenericRepository.java
```diff
@@ -69,10 +69,10 @@ public void deleteInBatch(final Iterable<E> entities) {
     @Override
     public List<E> findAll() {
         try {
-            return this.graph.traversal().V().has(ZONE_ID_KEY).has(getEntityIdKey()).toList().stream()
+            return this.graphTraversal.V().has(ZONE_ID_KEY).has(getEntityIdKey()).toList().stream()
                     .map(this::vertexToEntity).collect(Collectors.toList());
         } finally {
-            this.graph.tx().commit();
+            this.graphTraversal.tx().commit();
```
anurag-gujral, line 75: Since we are doing a findAll and not modifying the graph I am not sure we need to call this.graphTraversal.tx().commit();

anurag-gujral (commented):
Let us discuss this next week and then we will go from there , on whether we need to call commit and rollback at all the places it is presently being called.

service/src/main/java/com/ge/predix/acs/privilege/management/dao/GraphGenericRepository.java
Cruppelt: I spoke with Sanjeev and Anubhav, and they want to keep it for now for historical reasons.

service/src/main/java/com/ge/predix/acs/privilege/management/dao/GraphGenericRepository.java
```diff
@@ -112,9 +112,9 @@ public E getOne(final Long id) {
         List<S> savedEntities = new ArrayList<>();
         try {
```
anubhavi25: This try-catch template:
```
try {
    ...
    this.graphTraversal.tx().commit();
} catch (Exception e) {
    this.graphTraversal.tx().rollback();
    throw e;
}
```
is used in so many places that it's worth refactoring it out into a function and using that everywhere instead, something like so:
```
private void commitTransaction(final Runnable graphQuery) {
    try {
        graphQuery.run();
        this.graphTraversal.tx().commit();
    } catch (final Exception e) {
        this.graphTraversal.tx().rollback();
        throw e;
    }
}
```
and using it with a lambda like so:
```
this.commitTransaction(() -> this.graphTraversal.<some_function>);
```

service/src/test/java/com/ge/predix/acs/privilege/management/dao/GraphResourceRepositoryTest.java
```diff
@@ -75,28 +74,31 @@ public void teardown() {
     }
 
     private void setupTitanGraph() throws InterruptedException, ExecutionException {
-        this.graph = TitanFactory.build().set("storage.backend", "inmemory").open();
-        GraphConfig.createSchemaElements(this.graph);
+        Graph graph = TitanFactory.build().set("storage.backend", "inmemory").open();
+        GraphConfig.createSchemaElements(graph);
+        graph.traversal().V().drop().iterate();
```
anubhavi25: Would the following not work?

```
this.graphTraversalSource = graph.traversal();
this.dropAllResources();
```

anubhavi25 (changes requested):

service/src/main/java/com/ge/predix/acs/privilege/management/dao/GraphGenericRepository.java
anubhavi25: @anurag-gujral: @sanjeevchopra mentioned that there might be nuances to the calls in the `try` block that warrant committing the transaction. In a future story, we can investigate removing the `finally` blocks everywhere this pattern occurs and ensure concurrency-specific tests pass.

anubhavi25 (approved):

Cruppelt:
@anurag-gujral We believe we need the commits and rollbacks. Documentation: http://tinkerpop.apache.org/docs/3.2.3/reference/#transactions
http://s3.thinkaurelius.com/docs/titan/1.0.0/tx.html

PR #134, opened 2017-07-07
US143577: Review and Investigate High Severity Issues for ACS

FrankGasparovic:
Signed-off-by: Chris <chris.ruppelt@ge.com>

### DIFF
--- a/commons/src/main/java/com/ge/predix/acs/commons/web/BaseRestApi.java
+++ b/commons/src/main/java/com/ge/predix/acs/commons/web/BaseRestApi.java
@@ -1,28 +1,23 @@
 /*******************************************************************************
- * Copyright 2016 General Electric Company. 
+ * Copyright 2016 General Electric Company.
  *
- * Licensed under the Apache License, Version 2.0 (the "License"); 
- * you may not use this file except in compliance with the License. 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
  * http://www.apache.org/licenses/LICENSE-2.0
  *
- * Unless required by applicable law or agreed to in writing, software 
- * distributed under the License is distributed on an "AS IS" BASIS, 
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. 
- * See the License for the specific language governing permissions and 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


service/src/main/java/com/ge/predix/acs/privilege/management/PrivilegeManagementException.java
```diff
@@ -62,4 +67,11 @@ public PrivilegeManagementException(final Throwable cause) {
         super(cause);
     }
 
+    public static ResponseEntity<JSONObject> handleIncorrectParamType() {
+        JSONObject response = new JSONObject();
+        response.put("error", "Bad Request");
+        response.put("message", "Request Parameter includeInheritedAttributes must be a boolean value");
```
anubhavi25: Pull out `"message"` into a utility class since it will definitely be used in the future (maybe somewhere in the `com.ge.predix.acs.privilege.management` package). Also `includeInheritedAttributes` is used in other places so it might be worth pulling it out into the same class.

service/src/main/java/com/ge/predix/acs/privilege/management/ResourcePrivilegeManagementController.java
```diff
@@ -204,4 +207,9 @@ private void validResourceIdentifierOrFail(final BaseResource resource, final St
         }
     }
 
+    @ExceptionHandler(MethodArgumentTypeMismatchException.class)
```
anubhavi25: To avoid duplicating this in multiple controller classes, consider including this in a `@ControllerAdvice`-annotated class with the `basePackages` or `annotations` selectors to narrow the subset of targeted controllers (also see: https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/bind/annotation/ControllerAdvice.html, http://blog.codeleak.pl/2013/11/controlleradvice-improvements-in-spring.html)

service/src/main/java/com/ge/predix/acs/privilege/management/PrivilegeManagementException.java
```diff
@@ -62,4 +67,11 @@ public PrivilegeManagementException(final Throwable cause) {
         super(cause);
     }
 
+    public static ResponseEntity<JSONObject> handleIncorrectParamType() {
+        JSONObject response = new JSONObject();
+        response.put("error", "Bad Request");
```
anubhavi25: Pull out `"error"` into a utility class since it will definitely be used in the future (maybe somewhere in the `com.ge.predix.acs.privilege.management` package). Also use `HttpStatus.BAD_REQUEST.getReasonPhrase()` instead of the `Bad Request` literal.

anubhavi25 (changes requested):

anubhavi25 (approved):

anubhavi25:
@FrankGasparovic @Cruppelt looks good 👍  merge when ready

PR #138, opened 2017-07-17
US144737: Enable audit service with audit sdk version 0.0.9 in CF3

FrankGasparovic:


irinaepshteyn:
@FrankGasparovic, please explain why integration tests are disabled

FrankGasparovic:
There are a few things outstanding that we need to discuss with audit team:

1. Integration tests are failing due to SSLValidation error. Need to figure out why that is happening and reenable integration test.
2. Some messages do not get audited because of the 1024 character limit in payloads. Some ACS payloads are larger then that.
3. We should consider in the future adding our PredixAuditIT to acceptance tests so that we verify Audit is working in deployed environment.

PR #155, opened 2017-09-11
Us140995 acs int test refactor Do not merge yet!!!!!!!!!!

sebastiantb:
Do not merge yet, this is for reviewing purposes only!


anubhavi25:
Good start 👍  . Initial comments have been discussed with @sebastiantb and @VoollaSandeepKumar (mostly around pulling out common zone-related logic from `ACSITSetUpFactory` into `ZoneFactory` and attempting to keep factory subclasses as stateless as possible i.e. they ideally should not include member variables that are not environment-specific `@Value`s).

PR #157, opened 2017-09-14
US159762: Updated copyright notices for all source files

anubhavi25:


anubhavi25:
Merging this in since there are no functional changes (builds have passed nonetheless)

PR #160, opened 2017-10-02
US174046: Updated the script that downloads third-party dependencies …

anubhavi25:
…to download transitive ones as well


anubhavi25:
There are no functional changes in this pull request

PR #164, opened 2017-10-09
DE50135: Fix the NullPointerException when attempting to set cache entry for an INDETERMINATE policy evaluation result

anubhavi25:


service/src/test/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCacheTest.java
```diff
@@ -102,8 +102,13 @@ public void testGetWithCacheMissForResource() {
         assertNull(this.cache.get(key));
     }
 
-    @Test
-    public void testGetWithCacheHit() {
+    @DataProvider
+    private Object[][] effects() {
```
irinaepshteyn: @anubhavi25, wouldn't it better to have a provider of not just effects but the whole policy evaluation results:

return new Object[][] { new Object[] { mockPermitResult() }, { mockIndeterminateResult() } };

This way we will avoid the if statement in line 124-125 below.

model/src/main/java/com/ge/predix/acs/rest/PolicyEvaluationResult.java
```diff
@@ -52,7 +52,7 @@
                     + "    /v1/site/{site_id}/plant/asset/{asset_id}\n" + "would include:\n" + "    /site/1\n"
```
irinaepshteyn: @anubhavi25, this comment is incorrect.  That's a good opportunity to fix it.
Change "/v1/site/{site_id}/plant/asset/{asset_id}" to the following attribute uri template: "/v1/site/{site_id}/plant{attribute_uri}".  Then it will be resolved to "/asset/1"

irinaepshteyn (dismissed):
I am done reviewing.  Please address my comments.

service/src/test/java/com/ge/predix/acs/policy/evaluation/cache/AbstractPolicyEvaluationCacheTest.java
anubhavi25: Agreed fixed

model/src/main/java/com/ge/predix/acs/rest/PolicyEvaluationResult.java
anubhavi25: Good catch fixed

anubhavi25:
Passing integration pull request builds:

https://predix1.jenkins.build.ge.com/job/Predix-Security/job/ACS/job/ACS-PullRequest/job/acs-integration-pullrequest/969/
https://predix1.jenkins.build.ge.com/job/Predix-Security/job/ACS/job/ACS-PullRequest/job/acs-integration-pullrequest/970/
https://predix1.jenkins.build.ge.com/job/Predix-Security/job/ACS/job/ACS-PullRequest/job/acs-integration-pullrequest/971/
https://predix1.jenkins.build.ge.com/job/Predix-Security/job/ACS/job/ACS-PullRequest/job/acs-integration-pullrequest/972/

anubhavi25:
@irinaepshteyn incorporated your additional feedback on not caching `INDETERMINATE` decisions since most of the reasons when this would occur are transient in nature and usu.
 can be fixed (XACML 3.0 also documents something similar: http://docs.oasis-open.org/xacml/3.0/xacml-3.0-core-spec-os-en.html#_Toc325047158)

irinaepshteyn (approved):

PR #175, opened 2017-10-27
Us187198 obligations

sebastiantb:
DO not merge yet, will be merged with evaluation story

### DIFF
--- a/model/src/main/java/com/ge/predix/acs/model/ActionArgument.java
+++ b/model/src/main/java/com/ge/predix/acs/model/ActionArgument.java
@@ -1,0 +1,105 @@
+/*******************************************************************************
+ * Copyright 2017 General Electric Company
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+
+package com.ge.predix.acs.model;
+
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+/**
+ * @author Sebastian Torres Brown
+ * 
+ *         An argument that used to resolve dynamic expressions in obligation action template.
+ *
+ */
+
+@ApiModel(description = "An argument that used to resolve dynamic expressions in obligation action template.")
+public class ActionArgument {
+
+    private String name;
+
+    private String value;
+
+    /**
+     * @return the name
+     */
+    @ApiModelProperty(required = true)
+    public String getName() {
+        return this.name;
+    }
+
+    /**
+     * @param name
+     *            the name to set
+     */
+    public void setName(final String name) {
+        this.name = name;
+    }
+
+    /**
+     * @return the value
+     */
+    @ApiModelProperty(required = true)
+    public String getValue() {
+        return this.value;
+    }
+
+    /**
+     * @param value
+     *            the value to set
+     */
+    public void setValue(final String value) {
+        this.value = value;
+    }
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + ((this.name == null) ? 0 : this.name.hashCode());
+        result = prime * result + ((this.value == null) ? 0 : this.value.hashCode());
+        return result;
+    }
+
+    @Override
+    public boolean equals(final Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        ActionArgument other = (ActionArgument) obj;
+        if (this.name == null) {
+            if (other.name != null) {
+                return false;
+            }
+        } else if (!this.name.equals(other.name)) {
+            return false;
+        }
+        if (this.value == null) {
+            if (other.value != null) {
+                return false;
+            }
+        } else if (!this.value.equals(other.value)) {
+            return false;
+        }
+        return true;
+    }
+}
--- a/model/src/main/java/com/ge/predix/acs/model/ObligationType.java
+++ b/model/src/main/java/com/ge/predix/acs/model/ObligationType.java
@@ -12,30 +12,18 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- *
- * SPDX-License-Identifier: Apache-2.0
  *******************************************************************************/
 
 package com.ge.predix.acs.model;
 
-import java.io.File;
-import java.io.IOException;
-
-import com.fasterxml.jackson.databind.ObjectMapper;
-
 /**
  *
- * @author acs-engineers@ge.com
+ * @author Sebastian Torres Brown
+ * 
+ * enum class to define the types of Obligations that are supported
+ * 
  */
-public final class PolicySets {
-    public static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
-    private PolicySets() {
-        // Prevents instantiation.
-    }
-
-    @SuppressWarnings("javadoc")
-    public static PolicySet loadFromFile(final File file) throws IOException {
-        return OBJECT_MAPPER.readValue(file, PolicySet.class);
-    }
+public enum ObligationType {
+    CUSTOM
 }
--- a/model/src/main/java/com/ge/predix/acs/model/Policy.java
+++ b/model/src/main/java/com/ge/predix/acs/model/Policy.java
@@ -34,6 +34,8 @@ public class Policy {
     private String name;
     private Target target;
     private List<Condition> conditions = Collections.emptyList();
+    private List<String> obligationIds = Collections.emptyList();
+
     private Effect effect;
 
     public String getName() {
@@ -69,6 +71,21 @@ public class Policy {
         this.conditions = conditions;
     }
 
+    /**
+     * @return the obligationIds
+     */
+    public List<String> getObligationIds() {
+        return this.obligationIds;
+    }
+
+    /**
+     * @param obligationIds
+     *            the obligationIds to set
+     */
+    public void setObligationIds(final List<String> obligationIds) {
+        this.obligationIds = obligationIds;
+    }
+
     @ApiModelProperty(required = true)
     public Effect getEffect() {
         return this.effect;
@@ -81,8 +98,8 @@ public class Policy {
     @SuppressWarnings("nls")
     @Override
     public String toString() {
-        return "Policy [name=" + this.name + ", target=" + this.target + ", conditions=" + this.conditions + ", effect="
-                + this.effect + "]";
+        return "Policy [name=" + this.name + ", target=" + this.target + ", conditions=" + this.conditions
+                + ", obligationIds=" + this.obligationIds + ", effect=" + this.effect + "]";
     }
 
 }
--- a/model/src/main/java/com/ge/predix/acs/model/PolicySet.java
+++ b/model/src/main/java/com/ge/predix/acs/model/PolicySet.java
@@ -19,6 +19,7 @@
 package com.ge.predix.acs.model;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 import io.swagger.annotations.ApiModel;
@@ -27,11 +28,14 @@ import io.swagger.annotations.ApiModelProperty;
 /**
  * @author acs-engineers@ge.com
  */
-@ApiModel(description = "A collection of access control policies evaluated in order. The first applicable policy"
-        + " determines the access control effect.")
+@ApiModel(
+        description = "A collection of access control policies evaluated in order. The first applicable policy"
+                + " determines the access control effect.")
 @SuppressWarnings("javadoc")
 public class PolicySet {
     private List<Policy> policies = new ArrayList<>();
+    private List<ObligationExpression> obligationExpressions = Collections.emptyList();
+
     private String name;
 
     public PolicySet() {
@@ -45,21 +49,35 @@ public class PolicySet {
     /**
      * @return All policies in this policy set.
      */
-    @ApiModelProperty(value = "A non empty list of Policies that define the Policy set",
-            required = true)
+    @ApiModelProperty(value = "A non empty list of Policies that define the Policy set", required = true)
     public List<Policy> getPolicies() {
         return this.policies;
     }
 
     /**
-     * @param policies the policies to set
+     * @param policies
+     *            the policies to set
      */
     public void setPolicies(final List<Policy> policies) {
         this.policies = policies;
     }
 
-    @ApiModelProperty(value = "User defined name for the Policy set",
-            required = false)
+    /**
+     * @return the obligationExpressions
+     */
+    public List<ObligationExpression> getObligationExpressions() {
+        return this.obligationExpressions;
+    }
+
+    /**
+     * @param obligationExpressions
+     *            the obligationExpressions to set
+     */
+    public void setObligationExpressions(final List<ObligationExpression> obligationExpressions) {
+        this.obligationExpressions = obligationExpressions;
+    }
+
+    @ApiModelProperty(value = "User defined name for the Policy set", required = false)
     public String getName() {
         return this.name;
     }
@@ -71,7 +89,8 @@ public class PolicySet {
     @SuppressWarnings("nls")
     @Override
     public String toString() {
-        return "PolicySet [policies=" + this.policies + ", name=" + this.name + "]";
+        return "PolicySet [policies=" + this.policies + ", obligations=" + this.obligationExpressions + ", name="
+                + this.name + "]";
     }
 
 }
--- a/service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidationMessages.java
+++ b/service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidationMessages.java
@@ -1,0 +1,44 @@
+/*******************************************************************************
+ * Copyright 2017 General Electric Company
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+
+package com.ge.predix.acs.service.policy.validation;
+
+/**
+ * @author acs-engineers@ge.com
+ *
+ */
+public final class PolicySetValidationMessages {
+
+    private PolicySetValidationMessages() {
+    }
+
+    public static final String POLICY_VAL_FAILED = "Policy validation failed.";
+
+    public static final String OBL_VAL_FAILED = "Obligation Expression validation failed.";
+
+    public static final String EXCEP_OBL_ACTION_ARG_DUPLICATED = OBL_VAL_FAILED
+            + " Names of actionArguments have to be unique."
+            + " Obligation expression [%s] contains the following repeated actionArguments: [%s].";
+
+    public static final String EXCEP_OBL_ACTION_TEMPLATE_NULL_EMPTY = OBL_VAL_FAILED
+            + " actionTemplate in obligation expression [%s] cannot be null or empty.";
+
+    public static final String EXCEP_OBL_EXPRESSIONS_DUPLICATED = OBL_VAL_FAILED
+            + " Id of obligation expression has to be unique. The following ids are repeated:  [%s]";
+
+    public static final String EXCEP_POLICY_OBL_IDS_NOT_FOUND = POLICY_VAL_FAILED
+            + " Unable to find matching obligation expression for the following obligationsIds in policy [%s]: [%s].";
+}
--- a/service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
+++ b/service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
@@ -27,6 +27,7 @@ import java.util.Set;
 
 import javax.annotation.PostConstruct;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -37,7 +38,9 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.ge.predix.acs.commons.policy.condition.ConditionScript;
 import com.ge.predix.acs.commons.policy.condition.groovy.GroovyConditionCache;
 import com.ge.predix.acs.commons.policy.condition.groovy.GroovyConditionShell;
+import com.ge.predix.acs.model.ActionArgument;
 import com.ge.predix.acs.model.Condition;
+import com.ge.predix.acs.model.ObligationExpression;
 import com.ge.predix.acs.model.Policy;
 import com.ge.predix.acs.model.PolicySet;
 import com.ge.predix.acs.utils.JsonUtils;
@@ -85,9 +88,59 @@ public class PolicySetValidatorImpl implements PolicySetValidator {
     @Override
     public void validatePolicySet(final PolicySet policySet) {
         validateSchema(policySet);
+        List<String> availableObligationIds = validateObligationsAndGetIds(policySet.getObligationExpressions());
         for (Policy p : policySet.getPolicies()) {
             validatePolicyConditions(p.getConditions());
             validatePolicyActions(p);
+            vaLidatePolicyObligations(availableObligationIds, p);
+        }
+    }
+
+    private List<String> validateObligationsAndGetIds(final List<ObligationExpression> obligationExps) {
+        List<String> ids = new ArrayList<>();
+        List<String> repeatedIds = new ArrayList<>();
+
+        for (ObligationExpression obligationExp : obligationExps) {
+            
+            if (ids.contains(obligationExp.getId())) {
+                repeatedIds.add(obligationExp.getId());
+            }
+
+            JsonNode node = JSONUTILS.readJsonNodeFromObject(obligationExp.getActionTemplate());
+            if (!node.fields().hasNext()) {
+                throw new PolicySetValidationException(String.format(
+                        PolicySetValidationMessages.EXCEP_OBL_ACTION_TEMPLATE_NULL_EMPTY, obligationExp.getId()));
+            }
+            
+            validateObligationsActionArguments(obligationExp.getId(), obligationExp.getActionArguments());
+            ids.add(obligationExp.getId());
+        }
+        if (!CollectionUtils.isEmpty(repeatedIds)) {
+            throw new PolicySetValidationException(
+                    String.format(PolicySetValidationMessages.EXCEP_OBL_EXPRESSIONS_DUPLICATED,
+                            Arrays.toString(repeatedIds.toArray())));
+        }
+        return ids;
+    }
+
+    private void validateObligationsActionArguments(final String obligationId,
+            final List<ActionArgument> actionArguments) {
+        if (CollectionUtils.isEmpty(actionArguments)) {
+            return;
+        }
+
+        List<String> iteratedActionArguments = new ArrayList<>();
+        List<String> duplicatedActionArguments = new ArrayList<>();
+        for (ActionArgument actionArgument : actionArguments) {
+            if (iteratedActionArguments.contains(actionArgument.getName())) {
+                duplicatedActionArguments.add(actionArgument.getName());
+            }
+            iteratedActionArguments.add(actionArgument.getName());
+        }
+        if (!CollectionUtils.isEmpty(duplicatedActionArguments)) {
+            throw new PolicySetValidationException(
+                    String.format(PolicySetValidationMessages.EXCEP_OBL_ACTION_ARG_DUPLICATED, obligationId,
+                            Arrays.toString(duplicatedActionArguments.toArray())));
         }
     }
 
@@ -104,11 +157,31 @@ public class PolicySetValidatorImpl implements PolicySetValidator {
             }
             for (String action : policyActions.split("\\s*,\\s*")) {
                 if (!this.validAcsPolicyHttpActionsSet.contains(action)) {
-                    throw new PolicySetValidationException(String.format("Policy Action validation failed: "
-                                    + "the action: [%s] is not contained in the allowed set of actions: [%s]", action,
-                            this.validAcsPolicyHttpActions));
+                    throw new PolicySetValidationException(String.format(
+                            "Policy Action validation failed: "
+                                    + "the action: [%s] is not contained in the allowed set of actions: [%s]",
+                            action, this.validAcsPolicyHttpActions));
                 }
             }
+        }
+    }
+
+    private void vaLidatePolicyObligations(final List<String> availableObligationIds, final Policy policy) {
+        List<String> policyObligationIds = policy.getObligationIds();
+        if (CollectionUtils.isEmpty(policyObligationIds)) {
+            return;
+        }
+
+        List<String> notFound = new ArrayList<>();
+        for (String obligationId : policyObligationIds) {
+            if (!availableObligationIds.contains(obligationId)) {
+                notFound.add(obligationId);
+            }
+        }
+        if (!notFound.isEmpty()) {
+            throw new PolicySetValidationException(
+                    String.format(PolicySetValidationMessages.EXCEP_POLICY_OBL_IDS_NOT_FOUND, policy.getName(),
+                            Arrays.toString(notFound.toArray())));
         }
     }
 
@@ -142,7 +215,7 @@ public class PolicySetValidatorImpl implements PolicySetValidator {
     public List<ConditionScript> validatePolicyConditions(final List<Condition> conditions) {
         List<ConditionScript> conditionScripts = new ArrayList<>();
         try {
-            if ((conditions == null) || conditions.isEmpty()) {
+            if (CollectionUtils.isEmpty(conditions)) {
                 return conditionScripts;
             }
             for (Condition condition : conditions) {
--- a/service/src/test/java/com/ge/predix/acs/service/policy/admin/PolicyManagementServiceTest.java
+++ b/service/src/test/java/com/ge/predix/acs/service/policy/admin/PolicyManagementServiceTest.java
@@ -66,12 +66,13 @@ import com.ge.predix.acs.zone.resolver.ZoneResolver;
 @Test
 @TestPropertySource("classpath:application.properties")
 @ActiveProfiles(resolver = TestActiveProfilesResolver.class)
-@ContextConfiguration(classes = { InMemoryPolicyEvaluationCache.class, InMemoryDataSourceConfig.class,
-        AttributeCacheFactory.class, PolicyManagementServiceImpl.class, SpringSecurityPolicyContextResolver.class,
-        PolicySetValidatorImpl.class, GroovyConditionShell.class, SpringSecurityZoneResolver.class,
-        GroovyConditionCache.class, AttributeConnectorServiceImpl.class, AttributeReaderFactory.class,
-        PrivilegeServiceResourceAttributeReader.class, PrivilegeServiceSubjectAttributeReader.class,
-        PrivilegeManagementServiceImpl.class, SubjectRepositoryProxy.class, ResourceRepositoryProxy.class })
+@ContextConfiguration(
+        classes = { InMemoryPolicyEvaluationCache.class, InMemoryDataSourceConfig.class, AttributeCacheFactory.class,
+                PolicyManagementServiceImpl.class, SpringSecurityPolicyContextResolver.class,
+                PolicySetValidatorImpl.class, GroovyConditionShell.class, SpringSecurityZoneResolver.class,
+                GroovyConditionCache.class, AttributeConnectorServiceImpl.class, AttributeReaderFactory.class,
+                PrivilegeServiceResourceAttributeReader.class, PrivilegeServiceSubjectAttributeReader.class,
+                PrivilegeManagementServiceImpl.class, SubjectRepositoryProxy.class, ResourceRepositoryProxy.class })
 public class PolicyManagementServiceTest extends AbstractTransactionalTestNGSpringContextTests {
 
     private static final String SUBDOMAIN1 = "tenant1";
@@ -96,8 +97,8 @@ public class PolicyManagementServiceTest extends AbstractTransactionalTestNGSpri
 
     private final ZoneEntity zone1 = this.createZone("zone1", SUBDOMAIN1, "description for Zone1");
     private final ZoneEntity zone2 = this.createZone("zone2", SUBDOMAIN2, "description for Zone2");
-    private final ZoneEntity defaultZone = this
-            .createZone("defaultZone", DEFAULT_SUBDOMAIN, "description for defaultZone");
+    private final ZoneEntity defaultZone = this.createZone("defaultZone", DEFAULT_SUBDOMAIN,
+            "description for defaultZone");
 
     @BeforeClass
     public void beforeClass() {
@@ -154,8 +155,8 @@ public class PolicyManagementServiceTest extends AbstractTransactionalTestNGSpri
 
     @Test
     public void testCreateApmPolicySetPositive() {
-        PolicySet policySet = this.jsonUtils
-                .deserializeFromFile("testApmPolicySetLoadsSuccessfully.json", PolicySet.class);
+        PolicySet policySet = this.jsonUtils.deserializeFromFile("testApmPolicySetLoadsSuccessfully.json",
+                PolicySet.class);
         String policyName = policySet.getName();
         this.policyService.upsertPolicySet(policySet);
         PolicySet savedPolicySet = this.policyService.getPolicySet(policyName);
@@ -163,6 +164,27 @@ public class PolicyManagementServiceTest extends AbstractTransactionalTestNGSpri
         Assert.assertEquals(savedPolicySet.getName(), policyName);
         this.policyService.deletePolicySet(policyName);
         Assert.assertEquals(this.policyService.getAllPolicySets().size(), 0);
+    }
+
+    @Test
+    public void testCreatePolicySetWithObligations() {
+        PolicySet policySet = this.jsonUtils.deserializeFromFile("obligation/set-with-obligation-expressions.json",
+                PolicySet.class);
+        String policyName = policySet.getName();
+        this.policyService.upsertPolicySet(policySet);
+        PolicySet savedPolicySet = this.policyService.getPolicySet(policyName);
+        Assert.assertNotNull(savedPolicySet);
+        Assert.assertEquals(this.jsonUtils.readJsonNodeFromObject(policySet).toString(),
+                this.jsonUtils.readJsonNodeFromObject(savedPolicySet).toString());
+        this.policyService.deletePolicySet(policyName);
+        Assert.assertEquals(this.policyService.getAllPolicySets().size(), 0);
+    }
+
+    @Test(expectedExceptions = { PolicyManagementException.class })
+    public void testCreatePolicySetWithInvalidObligations() {
+        PolicySet policySet = this.jsonUtils.deserializeFromFile("obligation/obligation-expression-id-not-unique.json",
+                PolicySet.class);
+        this.policyService.upsertPolicySet(policySet);
     }
 
     public void testUpdatePolicySet() {
@@ -190,8 +212,8 @@ public class PolicyManagementServiceTest extends AbstractTransactionalTestNGSpri
 
     public void testCreateMultiplePolicySets() {
         PolicySet policySet = this.jsonUtils.deserializeFromFile("set-with-1-policy.json", PolicySet.class);
-        PolicySet policySet2 = this.jsonUtils
-                .deserializeFromFile("policy-set-with-one-policy-one-condition.json", PolicySet.class);
+        PolicySet policySet2 = this.jsonUtils.deserializeFromFile("policy-set-with-one-policy-one-condition.json",
+                PolicySet.class);
 
         this.policyService.upsertPolicySet(policySet);
         try {
@@ -209,15 +231,15 @@ public class PolicyManagementServiceTest extends AbstractTransactionalTestNGSpri
 
     @Test(expectedExceptions = { PolicyManagementException.class })
     public void testCreatePolicySetWithInvalidConditions() {
-        PolicySet policySet = this.jsonUtils
-                .deserializeFromFile("policy-set-with-one-policy-invalid-condition.json", PolicySet.class);
+        PolicySet policySet = this.jsonUtils.deserializeFromFile("policy-set-with-one-policy-invalid-condition.json",
+                PolicySet.class);
         this.policyService.upsertPolicySet(policySet);
     }
 
     @Test(expectedExceptions = { PolicyManagementException.class })
     public void testCreatePolicySetWithInvalidJson() {
-        PolicySet policySet = this.jsonUtils
-                .deserializeFromFile("policyset/validator/test/missing-effect-policy.json", PolicySet.class);
+        PolicySet policySet = this.jsonUtils.deserializeFromFile("policyset/validator/test/missing-effect-policy.json",
+                PolicySet.class);
         this.policyService.upsertPolicySet(policySet);
     }
 
@@ -235,8 +257,8 @@ public class PolicyManagementServiceTest extends AbstractTransactionalTestNGSpri
 
     public void testCreatePolicySetsForMultipleApplications() {
         PolicySet client1PolicySet = this.jsonUtils.deserializeFromFile("set-with-1-policy.json", PolicySet.class);
-        PolicySet client2PolicySet = this.jsonUtils
-                .deserializeFromFile("policy-set-with-one-policy-one-condition.json", PolicySet.class);
+        PolicySet client2PolicySet = this.jsonUtils.deserializeFromFile("policy-set-with-one-policy-one-condition.json",
+                PolicySet.class);
 
         Mockito.when(this.mockZoneResolver.getZoneEntityOrFail()).thenReturn(this.zone1);
 
@@ -271,8 +293,8 @@ public class PolicyManagementServiceTest extends AbstractTransactionalTestNGSpri
 
     public void testCreatePolicySetsForMultipleZones() {
         PolicySet issuer1PolicySet = this.jsonUtils.deserializeFromFile("set-with-1-policy.json", PolicySet.class);
-        PolicySet issuer2PolicySet = this.jsonUtils
-                .deserializeFromFile("policy-set-with-one-policy-one-condition.json", PolicySet.class);
+        PolicySet issuer2PolicySet = this.jsonUtils.deserializeFromFile("policy-set-with-one-policy-one-condition.json",
+                PolicySet.class);
 
         Mockito.when(this.mockZoneResolver.getZoneEntityOrFail()).thenReturn(this.zone1);
         this.policyService.upsertPolicySet(issuer1PolicySet);
--- a/service/src/test/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorTest.java
+++ b/service/src/test/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorTest.java
@@ -50,7 +50,7 @@ import com.ge.predix.acs.utils.JsonUtils;
         classes = { GroovyConditionCache.class, GroovyConditionShell.class, PolicySetValidatorImpl.class })
 public class PolicySetValidatorTest extends AbstractTestNGSpringContextTests {
 
-    private final JsonUtils jsonUtils = new JsonUtils();
+    private static JsonUtils jsonUtils = new JsonUtils();
     @Autowired
     private PolicySetValidator policySetValidator;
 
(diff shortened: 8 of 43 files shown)


model/src/main/java/com/ge/predix/acs/model/Obligation.java
```diff
@@ -0,0 +1,97 @@
+package com.ge.predix.acs.model;
```
VoollaSandeepKumar: Do not forget to add/(run the script) the License for newly created files

model/src/main/java/com/ge/predix/acs/model/ActionArgument.java
```diff
@@ -0,0 +1,51 @@
+package com.ge.predix.acs.model;
```
VoollaSandeepKumar, line 17: License add/(run the script)

model/src/main/java/com/ge/predix/acs/model/Obligation.java
```diff
@@ -0,0 +1,97 @@
...
+
```
VoollaSandeepKumar: package is missing and License add/(run the script)

VoollaSandeepKumar (commented):
Basic review comments. Overall looks good .

service/src/test/java/com/ge/predix/acs/service/policy/admin/PolicyManagementServiceTest.java
```diff
@@ -163,6 +164,113 @@ public void testCreateApmPolicySetPositive() {
...
+            expectedExceptions = { PolicyManagementException.class },
+            expectedExceptionsMessageRegExp = ".*Obligation validation failed\\. actionTemplate"
+                    + " cannot be null or empty on index.*")
+    public void testValidatePolicyBadObligationActionTemplateBlank() {
+        PolicySet policySet = this.jsonUtils
+                .deserializeFromFile("obligation/set-with-1-policy-bad-action-template-blank.json", PolicySet.class);
+        this.policyService.upsertPolicySet(policySet);
+    }
+    
+    @Test(
```
VoollaSandeepKumar: Are these test cases tested before disabling them with (enabled = false) and what is the reason for disabling them ?

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -71,6 +77,31 @@
 
     private Set<String> validAcsPolicyHttpActionsSet;
 
+    private static final String EXCEP_POLICY_OBL_IDS_NULL_EMPTY = "Policy validation failed. obligationsIds cannot"
```
VoollaSandeepKumar: It would be good if we can move all this error messages to existing utility file or creating a new one

model/src/main/java/com/ge/predix/acs/model/Policy.java
```diff
@@ -79,8 +96,8 @@ public void setEffect(final Effect effect) {
     @SuppressWarnings("nls")
     @Override
     public String toString() {
-        return "Policy [name=" + this.name + ", target=" + this.target + ", conditions=" + this.conditions + ", effect="
-                + this.effect + "]";
+        return "Policy [name=" + this.name + ", target=" + this.target + ", conditions=" + this.conditions
+                + ", obligationIds=" + this.obligationIds + ", effect=" + this.effect + "]";
```
VoollaSandeepKumar, line 102: Assigning "conditions=this.conditions" will print the reference/hash of the list. Question is do we need the reference/hash of the conditions or the actual values of the conditions. If we need the actual values this code needs to be changed.

model/src/main/java/com/ge/predix/acs/model/ActionArgument.java
sebastiantb: done

model/src/main/java/com/ge/predix/acs/model/Obligation.java
sebastiantb: done

model/src/main/java/com/ge/predix/acs/model/Obligation.java
sebastiantb: package? without it it won't compile lol

service/src/test/java/com/ge/predix/acs/service/policy/admin/PolicyManagementServiceTest.java
sebastiantb: forgot to push  the file with test enabled

model/src/main/java/com/ge/predix/acs/model/Policy.java
sebastiantb: decided not to proceed with this

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
sebastiantb: moved

VoollaSandeepKumar (approved):

model/src/main/java/com/ge/predix/acs/model/Obligation.java
```diff
@@ -0,0 +1,113 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+
+package com.ge.predix.acs.model;
+
+import java.util.Collections;
+import java.util.List;
+
+//import javax.validation.constraints.Max;
```
anubhavi25: Remove these commented-out imports if they're not needed

model/src/main/java/com/ge/predix/acs/model/Obligation.java
```diff
@@ -0,0 +1,113 @@
...
+                + "indicated Obligation(s)' or, alternatively: 'Deny, and in addition fulfill the indicated"
+                + " Obligation(s)'.  If the PEP does not understand, or cannot fulfill, any of the obligations, "
+                + "then it MUST act as if the PDP had returned a 'Deny' authorization decision value.")
+public class Obligation {
+
+    private String id;
+
+    private ObligationType type;
+
+    private Object actionTemplate;
```
anubhavi25: Since `actionTemplate` specifies arbitrary JSON, replace the generic `Object` type with the JSON-specific `JsonNode` type provided by Jackson so that's easily verifiable and parseable by validation-specific methods if necessary.

model/src/main/java/com/ge/predix/acs/model/PolicySet.java
```diff
@@ -25,11 +25,15 @@
...
-@ApiModel(description = "A collection of access control policies evaluated in order. The first applicable policy"
-        + " determines the access control effect.")
+@ApiModel(
+        description = "A collection of access control policies evaluated in order. The first applicable policy"
+                + " determines the access control effect.")
 @SuppressWarnings("javadoc")
 public class PolicySet {
     private List<Policy> policies = new ArrayList<>();
+
+    private List<Obligation> obligations = new ArrayList<>();
```
anubhavi25: Since we're conforming to XACML 3.0, this field will be deserialized from the request body as `obligationExpressions` and serialized to the response body as `obligations` so we need to add getters/setters appropriately that use `@JsonProperty` to switch the attribute name (see the design document and https://stackoverflow.com/a/8560390).

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidationMessages.java
```diff
@@ -0,0 +1,53 @@
...
+    public static final String EXCEP_POLICY_OBL_IDS_NULL_EMPTY = "Policy validation failed. obligationsIds cannot"
+            + " contain null or empty values for policy: [%s]";
+
+    public static final String EXCEP_OBL_IDS_REPEATED = "Obligation validation failed. Obligation ids need to unique."
+            + " Repeated values:  %s";
+
+    public static final String EXCEP_OBL_ACTION_ARG_NULL_BLANK = "Obligation validation failed. actionArgument [%s]  "
+            + "cannot be null or blank on [index: [%d] , actionArgument index: [%d] ]";
+
+    public static final String EXCEP_OBL_ACTION_ARG_NULL = "Obligation validation failed. actionArgument [%s]  "
```
anubhavi25: Remove this variable since its message is already included in the `EXCEP_OBL_ACTION_ARG_NULL_BLANK` variable above it (i.e. `..cannot be null or blank...` is already a substring of the previous message).

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidationMessages.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * @author acs-engineers@ge.com
+ *
+ */
+public final class PolicySetValidationMessages {
+
+    private PolicySetValidationMessages() {
+    }
+
+    public static final String EXCEP_POLICY_OBL_IDS_NULL_EMPTY = "Policy validation failed. obligationsIds cannot"
```
anubhavi25: Keep DRY by pulling the `Policy validation failed` substring out to a `static final` variable and use that everywhere instead.

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidationMessages.java
```diff
@@ -0,0 +1,53 @@
...
+            + " contain null or empty values for policy: [%s]";
+
+    public static final String EXCEP_OBL_IDS_REPEATED = "Obligation validation failed. Obligation ids need to unique."
```
anubhavi25: 1. Keep DRY by pulling the `Obligation validation failed` substring out to a `static final` variable and use that everywhere instead.
2. Typo: `need to` -> `need to be`

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -83,12 +89,73 @@ public void removeCachedConditions(final PolicySet policySet) {
...
         for (Policy p : policySet.getPolicies()) {
             validatePolicyConditions(p.getConditions());
             validatePolicyActions(p);
+            vaLidatePolicyObligations(obligationIds, p);
         }
     }
 
+    @SuppressWarnings("unchecked")
+    private List<String> validateObligationsAndGetIds(final List<Obligation> obligations) {
+        List<String> ids = new ArrayList<String>();
```
anubhavi25: Remove the redundant type from the RHS (the diamond operator `<>` is enough to correctly deduce the type).

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -83,12 +89,73 @@ public void removeCachedConditions(final PolicySet policySet) {
...
+        List<String> repeatedIds = new ArrayList<String>();
```
anubhavi25: Remove the redundant type from the RHS (the diamond operator `<>` is enough to correctly deduce the type).

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -83,12 +89,73 @@ public void removeCachedConditions(final PolicySet policySet) {
...
+        if (!CollectionUtils.isEmpty(repeatedIds)) {
+            throw new PolicySetValidationException(String.format(PolicySetValidationMessages.EXCEP_OBL_IDS_REPEATED,
+                    Arrays.toString(repeatedIds.toArray())));
+        }
+        return ids;
+    }
+
+    private void validateActionArguments(final int obligationIndex, final List<ActionArgument> actionArguments) {
+        AtomicInteger index = new AtomicInteger(0);
+        List<String> iteratedActionArguments = new ArrayList<String>();
```
anubhavi25: Remove the redundant type from the RHS (the diamond operator `<>` is enough to correctly deduce the type).

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -83,12 +89,73 @@ public void removeCachedConditions(final PolicySet policySet) {
...
+        List<String> duplicatedActionArguments = new ArrayList<String>();
```
anubhavi25: Remove the redundant type from the RHS (the diamond operator `<>` is enough to correctly deduce the type).

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidationMessages.java
```diff
@@ -0,0 +1,53 @@
...
+            + "cannot be repeated on  index : [%d], values: [%s]";
+
+    public static final String EXCEP_OBL_ID_NULL_BLANK = "Obligation validation failed. id cannot be null or blank "
+            + "on index: [%s]";
+
+    public static final String EXCEP_OBL_ACTION_TEMPLATE_NULL_BLANK = "Obligation validation failed. actionTemplate"
+            + " cannot be null or empty on index: [%s]";
+
+    public static final String EXCEP_POLICY_OBL_IDS_NOT_FOUND = "Policy validation failed. The following "
+            + "obligationsIds  does not match with the obligations defined for this policy set. [ policy: [%s]"
```
anubhavi25: Rephrase: `obligationsIds does not match with the` -> `obligationIds do not match the`

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidationMessages.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public static final String EXCEP_POLICY_OBL_IDS_NULL_EMPTY = "Policy validation failed. obligationsIds cannot"
+            + " contain null or empty values for policy: [%s]";
+
+    public static final String EXCEP_OBL_IDS_REPEATED = "Obligation validation failed. Obligation ids need to unique."
+            + " Repeated values:  %s";
+
+    public static final String EXCEP_OBL_ACTION_ARG_NULL_BLANK = "Obligation validation failed. actionArgument [%s]  "
+            + "cannot be null or blank on [index: [%d] , actionArgument index: [%d] ]";
```
anubhavi25: Some messages use `null or blank` while others use `null or empty`. I think we should consistently use the latter everywhere since it's clearer wrt JSON. Note that with this change the names of unit tests will also change as well.

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -83,12 +89,73 @@ public void removeCachedConditions(final PolicySet policySet) {
...
             validatePolicyActions(p);
+            vaLidatePolicyObligations(obligationIds, p);
         }
     }
 
+    @SuppressWarnings("unchecked")
+    private List<String> validateObligationsAndGetIds(final List<Obligation> obligations) {
+        List<String> ids = new ArrayList<String>();
+        List<String> repeatedIds = new ArrayList<String>();
+        AtomicInteger index = new AtomicInteger(0);
```
anubhavi25: Since `index` is a simple counter used in the `for` loop below and we're operating in the context of a single thread here, a simple `int` should suffice and would be more performant.

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -83,12 +89,73 @@ public void removeCachedConditions(final PolicySet policySet) {
...
+        }
+        if (!CollectionUtils.isEmpty(repeatedIds)) {
+            throw new PolicySetValidationException(String.format(PolicySetValidationMessages.EXCEP_OBL_IDS_REPEATED,
+                    Arrays.toString(repeatedIds.toArray())));
+        }
+        return ids;
+    }
+
+    private void validateActionArguments(final int obligationIndex, final List<ActionArgument> actionArguments) {
+        AtomicInteger index = new AtomicInteger(0);
```
anubhavi25: Since `index` is a simple counter used in the `for` loop below and we're operating in the context of a single thread here, a simple `int` should suffice and would be more performant.

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -83,12 +89,73 @@ public void removeCachedConditions(final PolicySet policySet) {
...
+            if (StringUtils.isEmpty(obligation.getId())) {
+                throw new PolicySetValidationException(
+                        String.format(PolicySetValidationMessages.EXCEP_OBL_ID_NULL_BLANK, index.get()));
+            }
+            if (ids.contains(obligation.getId())) {
+                repeatedIds.add(obligation.getId());
+            }
+
+            if (null == obligation.getActionTemplate()
+                    || ((Map<Object, Object>) obligation.getActionTemplate()).size() == 0) {
```
anubhavi25: Prefer `.isEmpty()` to `.size() == 0`

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -102,14 +169,35 @@ private void validatePolicyActions(final Policy p) {
...
+                            "Policy Action validation failed: "
+                                    + "the action: [%s] is not contained in the allowed set of actions: [%s]",
+                            action, this.validAcsPolicyHttpActions));
                 }
             }
         }
     }
 
+    private void vaLidatePolicyObligations(final List<String> obligationIds, final Policy policy) {
+        List<String> notFound = new ArrayList<String>();
```
anubhavi25: Remove the redundant type from the RHS (the diamond operator `<>` is enough to correctly deduce the type).

service/src/test/java/com/ge/predix/acs/service/policy/admin/PolicyManagementServiceTest.java
```diff
@@ -163,6 +164,106 @@ public void testCreateApmPolicySetPositive() {
...
+        PolicySet policySet = this.jsonUtils.deserializeFromFile("obligation/set-with-1-policy.json", PolicySet.class);
+        String policyName = policySet.getName();
+        this.policyService.upsertPolicySet(policySet);
+        PolicySet savedPolicySet = this.policyService.getPolicySet(policyName);
+        Assert.assertEquals(2, savedPolicySet.getObligations().size());
+        this.policyService.deletePolicySet(policyName);
+        Assert.assertEquals(this.policyService.getAllPolicySets().size(), 0);
+    }
+
+    @Test(
```
anubhavi25: This and the next `@Test` can be combined by parameterizing the file path and providing possible paths to test via a `@DataProvider`.

service/src/test/java/com/ge/predix/acs/service/policy/admin/PolicyManagementServiceTest.java
```diff
@@ -163,6 +164,106 @@ public void testCreateApmPolicySetPositive() {
...
+            expectedExceptions = { PolicyManagementException.class },
+            expectedExceptionsMessageRegExp = ".*Obligation validation failed\\. id cannot be null or blank on index.*")
+    public void testValidatePolicyBadObligationIdBlank() {
+        PolicySet policySet = this.jsonUtils
+                .deserializeFromFile("obligation/set-with-1-policy-bad-obligations-id-blank.json", PolicySet.class);
+        this.policyService.upsertPolicySet(policySet);
+    }
+
+    @Test(
```
anubhavi25: This and the next `@Test` can be combined by parameterizing the file path and providing possible paths to test via a `@DataProvider`.

service/src/test/java/com/ge/predix/acs/service/policy/admin/PolicyManagementServiceTest.java
```diff
@@ -163,6 +164,106 @@ public void testCreateApmPolicySetPositive() {
         Assert.assertEquals(this.policyService.getAllPolicySets().size(), 0);
     }
 
+    @Test
+    public void testCreatePolicyOkObligations() {
+        PolicySet policySet = this.jsonUtils.deserializeFromFile("obligation/set-with-1-policy.json", PolicySet.class);
+        String policyName = policySet.getName();
+        this.policyService.upsertPolicySet(policySet);
+        PolicySet savedPolicySet = this.policyService.getPolicySet(policyName);
+        Assert.assertEquals(2, savedPolicySet.getObligations().size());
```
anubhavi25: Asserting the contents of the returned obligations would make this a more robust test (and obviate the need for asserting the size).

service/src/test/java/com/ge/predix/acs/service/policy/admin/PolicyManagementServiceTest.java
```diff
@@ -163,6 +164,106 @@ public void testCreateApmPolicySetPositive() {
...
+        this.policyService.deletePolicySet(policyName);
+        Assert.assertEquals(this.policyService.getAllPolicySets().size(), 0);
+    }
+
+    @Test(
+            expectedExceptions = { PolicyManagementException.class },
+            expectedExceptionsMessageRegExp = ".*Obligation validation failed\\. id cannot be null or blank on index.*")
```
anubhavi25: This test and all tests below it are annotated to use the `expectedExceptions` and `expectedExceptionsMessageRegExp` attributes which make them perfect candidates for refactoring to use an [annotation transformer](https://stackoverflow.com/a/36247029) instead of duplicating code.

service/src/test/java/com/ge/predix/acs/service/policy/admin/PolicyManagementServiceTest.java
```diff
@@ -163,6 +164,106 @@ public void testCreateApmPolicySetPositive() {
...
+    @Test(
+            expectedExceptions = { PolicyManagementException.class },
+            expectedExceptionsMessageRegExp = ".*Obligation validation failed\\. Obligation ids need to unique\\."
+                    + " Repeated values.*")
+    public void testValidateObligationIdsNotUnique() {
+        PolicySet policySet = this.jsonUtils
+                .deserializeFromFile("obligation/set-with-1-policy-bad-obligation-ids-repeated.json", PolicySet.class);
+        this.policyService.upsertPolicySet(policySet);
+    }
+
```
anubhavi25: Add positive and negative tests for:
- Testing the validity of the `type` of obligation
- Testing the validity of the `optional` attribute
- Testing the validity of the `actionTemplate.select` list in the presence of duplicated `name`s (might be an edge case?)
- Testing the validity of the `actionTemplate.filter` attribute in the presence of placeholders that don't get expanded since they're absent from the `actionArguments` list

anubhavi25 (changes requested):

model/src/main/java/com/ge/predix/acs/model/Obligation.java
sebastiantb: added license

model/src/main/java/com/ge/predix/acs/model/Obligation.java
sebastiantb: done

service/src/test/java/com/ge/predix/acs/service/policy/admin/PolicyManagementServiceTest.java
sebastiantb: Type is enum, other values rather than  CUSTOM, WILL throw a 400
Optional has been removes.

Testing the validity of the actionTemplate.select list in the presence of duplicated names (might be an edge case?)
Don't understand this action template is any object, can contain a collection or not

Testing the validity of the actionTemplate.filter attribute in the presence of placeholders that don't get expanded since they're absent from the actionArguments list

Don't understand this action template is any object, can contain a collection or not

service/src/test/java/com/ge/predix/acs/service/policy/admin/PolicyManagementServiceTest.java
sebastiantb: I'll still have to use the new annotation and send the regex expression to match.
I don't think that annotations can be considered  as repeated code, since it's not newly codified functionality repeated several times.

model/src/main/java/com/ge/predix/acs/model/Obligation.java
```diff
@@ -0,0 +1,113 @@
...
+     *            the type to set
+     */
+    public void setType(final ObligationType type) {
+        this.type = type;
+    }
+
+    /**
+     * @return the actionTemplate
+     */
+    public Object getActionTemplate() {
```
sanjeevchopra: * is there a reason for this not to be String? At minimum, we need to document that this must be serialize-able to a string. 
* @team is the name `actionTemplate` from XACML - i dont see it. Otherwise, it may be better to chose something else to avoid any confusion with Policy.Target.Action.  how about `constraint`?

model/src/main/java/com/ge/predix/acs/model/ObligationType.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ *
+ * @author Sebastian Torres Brown
+ * 
+ * enum class to define the types of Obligations that are supported
+ * 
+ */
+
+public enum ObligationType {
```
sanjeevchopra, line 27: we could add this, if and when we add other types.  Especially because we don't see any use-case for it.

model/src/main/java/com/ge/predix/acs/model/Policy.java
```diff
@@ -67,6 +69,21 @@ public void setConditions(final List<Condition> conditions) {
         this.conditions = conditions;
     }
 
+    /**
+     * @return the obligationIds
+     */
+    public Set<String> getObligationIds() {
```
sanjeevchopra: did we discuss/decide if obligations have a expected order of processing for PDP?

model/src/main/java/com/ge/predix/acs/model/PolicySet.java
sanjeevchopra: are we conforming to XACML 3.0? the actionTemplate/arguments are not defined in XACML.

sanjeevchopra (commented):
please update version of acs-model

model/src/main/java/com/ge/predix/acs/model/Policy.java
sebastiantb: we decided to maintain the order of the obligations  as the execution order, but this can be further implement with parallelization, grouping  etc..

service/src/test/java/com/ge/predix/acs/service/policy/admin/PolicyManagementServiceTest.java
sebastiantb: Since action template is a free object, we don't know what to validate against, we can consider in the future a mechanism to upload a contract specification for the action template.

model/src/main/java/com/ge/predix/acs/model/ObligationType.java
sebastiantb: we will make this not mandatory, defaulting to custom

model/src/main/java/com/ge/predix/acs/model/Obligation.java
sebastiantb: jsonnode fails on the openjpa layer when trying to save,  the rest controller layer makes sure the serialization into object is valid, if you send an invalid json object it will throw an erro 400.

service/src/test/java/com/ge/predix/acs/service/policy/admin/PolicyManagementServiceTest.java
sebastiantb: double checking that we are testing that action arguments are not repeated.

model/src/main/java/com/ge/predix/acs/model/Obligation.java
sanjeevchopra: @sebastiantb proposed just 'template'.  I am ok with it since it's inside the 'obligations' object. anyone have a better suggestion?

model/src/main/java/com/ge/predix/acs/model/Policy.java
sanjeevchopra: in that case, use a concrete Set implementation which preserves order - LinkedHashSet .  see PolicyEvaluationRequestV1#policySetsEvaluationOrder also uses this type.

model/src/main/java/com/ge/predix/acs/model/ObligationType.java
sanjeevchopra: 👍

model/src/main/java/com/ge/predix/acs/model/Obligation.java
irinaepshteyn: @sanjeevchopra, @sebastiantb, initially, actionTemplate inside obligationExpression was resolved into action inside obligation.  If we change actionTemplate to just template what are we going to call resolved template?

sanjeevchopra: didn't realize that. I'm ok with the current name then, i guess it is scoped by 'obligation'

model/src/main/java/com/ge/predix/acs/model/Policy.java
irinaepshteyn: Why don't we use List as we do for maintaining order of the policies in the policy set.

model/src/main/java/com/ge/predix/acs/model/Obligation.java
```diff
@@ -0,0 +1,109 @@
...
+@ApiModel(
+        description = "Obligation augments the interface contract between the policy engine (PDP) and "
+                + "enforcement point (PEP).  In addition to the actual access decision of either 'Permit' or 'Deny', "
+                + "the policy engine may return some supplementary information to the enforcement point in the form"
+                + " of an Obligation. Obligations are a set of operations that must be performed by the PEP "
+                + "in conjunction with an authorization decision, e.g., 'Permit, and in addition fulfill the "
+                + "indicated Obligation(s)' or, alternatively: 'Deny, and in addition fulfill the indicated"
+                + " Obligation(s)'.  If the PEP does not understand, or cannot fulfill, any of the obligations, "
+                + "then it MUST act as if the PDP had returned a 'Deny' authorization decision value.")
+public class Obligation {
```
irinaepshteyn: Should we call this class ObligationExpression?  

The difference between ObligationExpression and Obligation is that former contains ActionTemplate and ActionArguments.  ObligationExpression are included in policy sets.  

Obligation in turn contains Action - resolved ActionTemplate by substituting ActionArguments.  Obligation is included in PolicyEvaluationResult.

The way it's defined right now Obligation JSON included in PolicyEvaluationResult will contain ActionTemplate and ActionArguments.

model/src/main/java/com/ge/predix/acs/model/Obligation.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * @param type
+     *            the type to set
+     */
+    public void setType(final ObligationType type) {
+        this.type = type;
+    }
+
```
irinaepshteyn: What did we decide on "optional" field that should indicate if the obligation fulfillment by PEP is mandatory or optional.  Maybe a better name is fulfillmentRequired.

model/src/main/java/com/ge/predix/acs/model/Obligation.java
```diff
@@ -0,0 +1,109 @@
...
+    private ObligationType type = ObligationType.CUSTOM;
+
+    private Object actionTemplate;
+
+    private List<ActionArgument> actionArguments = Collections.emptyList();
+
+    /**
+     * @return the id
+     */
+    public String getId() {
```
irinaepshteyn: Add @ApiModelProperty(required = true)

model/src/main/java/com/ge/predix/acs/model/Obligation.java
```diff
@@ -0,0 +1,109 @@
...
+     */
+    public Object getActionTemplate() {
+        return this.actionTemplate;
+    }
+
+    /**
+     * @param actionTemplate
+     *            the actionTemplate to set
+     */
+    public void setActionTemplate(final Object actionTemplate) {
```
irinaepshteyn: Add @ApiModelProperty(required = true)

model/src/main/java/com/ge/predix/acs/model/PolicySet.java
```diff
@@ -19,17 +19,23 @@
...
  */
-@ApiModel(description = "A collection of access control policies evaluated in order. The first applicable policy"
-        + " determines the access control effect.")
+@ApiModel(
+        description = "A collection of access control policies evaluated in order. The first applicable policy"
+                + " determines the access control effect.")
 @SuppressWarnings("javadoc")
 public class PolicySet {
     private List<Policy> policies = new ArrayList<>();
+    @JsonProperty("obligationExpressions")
```
irinaepshteyn: Add unit test (extend existing PolicySetTest?) to test serialization/deserialization of obligations

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -83,10 +87,70 @@ public void removeCachedConditions(final PolicySet policySet) {
...
+    }
+
+    @SuppressWarnings("unchecked")
+    private List<String> validateObligationsAndGetIds(final List<Obligation> obligations) {
+        List<String> ids = new ArrayList<>();
+        List<String> repeatedIds = new ArrayList<>();
+        int index = 0;
+
+        for (Obligation obligation : obligations) {
+            if (StringUtils.isEmpty(obligation.getId())) {
```
irinaepshteyn: That requirement can be defined as part of JSON schema (@JsonProperty(required=true)) and validated as part of JSON schema validation.  Also add a unit test.

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -83,10 +87,70 @@ public void removeCachedConditions(final PolicySet policySet) {
...
+                throw new PolicySetValidationException(
+                        String.format(PolicySetValidationMessages.EXCEP_OBL_ID_NULL_EMPTY, index));
+            }
+            if (ids.contains(obligation.getId())) {
+                repeatedIds.add(obligation.getId());
+            }
+
+            if (null == obligation.getActionTemplate()
```
irinaepshteyn: Also can be unloaded to JSON schema validation

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -83,10 +87,70 @@ public void removeCachedConditions(final PolicySet policySet) {
...
+                    || ((Map<Object, Object>) obligation.getActionTemplate()).isEmpty()) {
+                throw new PolicySetValidationException(
+                        String.format(PolicySetValidationMessages.EXCEP_OBL_ACTION_TEMPLATE_NULL_EMPTY, index));
+            }
+            validateActionArguments(index, obligation.getActionArguments());
+            ids.add(obligation.getId());
+            index++;
+        }
+        if (!CollectionUtils.isEmpty(repeatedIds)) {
```
irinaepshteyn: Do we want to error out on repeated obligations?  If so, it would be better to define obligation ids as an ordered set to avoid duplication.

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -83,10 +87,70 @@ public void removeCachedConditions(final PolicySet policySet) {
...
+        }
+        return ids;
+    }
+
+    private void validateActionArguments(final int obligationIndex, final List<ActionArgument> actionArguments) {
+        int index = 0;
+        List<String> iteratedActionArguments = new ArrayList<>();
+        List<String> duplicatedActionArguments = new ArrayList<>();
+        for (ActionArgument actionArgument : actionArguments) {
+            if (StringUtils.isEmpty(actionArgument.getName())) {
```
irinaepshteyn: Unload to JSON schema validation.

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -83,10 +87,70 @@ public void removeCachedConditions(final PolicySet policySet) {
...
+                throw new PolicySetValidationException(String.format(
+                        PolicySetValidationMessages.EXCEP_OBL_ACTION_ARG_NULL_EMPTY, "name", obligationIndex, index));
+            }
+            if (StringUtils.isEmpty(actionArgument.getValue())) {
```
irinaepshteyn: Unload to JSON schema validation.

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -83,10 +87,70 @@ public void removeCachedConditions(final PolicySet policySet) {
...
+                throw new PolicySetValidationException(String.format(
+                        PolicySetValidationMessages.EXCEP_OBL_ACTION_ARG_NULL_EMPTY, "value", obligationIndex, index));
+            }
+            if (iteratedActionArguments.contains(actionArgument.getName())) {
+                duplicatedActionArguments.add(actionArgument.getName());
+            }
+            iteratedActionArguments.add(actionArgument.getName());
+            index++;
+        }
+        if (!CollectionUtils.isEmpty(duplicatedActionArguments)) {
```
irinaepshteyn: Duplicates could be excluded by using ordered Set instead of List.

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -83,10 +87,70 @@ public void removeCachedConditions(final PolicySet policySet) {
     @Override
     public void validatePolicySet(final PolicySet policySet) {
         validateSchema(policySet);
+        List<String> obligationIds = validateObligationsAndGetIds(policySet.getObligations());
```
irinaepshteyn: Please add unit tests for all new validation functionality

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -102,14 +166,35 @@ private void validatePolicyActions(final Policy p) {
...
+                    throw new PolicySetValidationException(String.format(
+                            "Policy Action validation failed: "
+                                    + "the action: [%s] is not contained in the allowed set of actions: [%s]",
+                            action, this.validAcsPolicyHttpActions));
                 }
             }
         }
     }
 
+    private void vaLidatePolicyObligations(final List<String> obligationIds, final Policy policy) {
```
irinaepshteyn: Please add a unit test

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -83,10 +87,70 @@ public void removeCachedConditions(final PolicySet policySet) {
...
+            index++;
+        }
+        if (!CollectionUtils.isEmpty(repeatedIds)) {
+            throw new PolicySetValidationException(String.format(PolicySetValidationMessages.EXCEP_OBL_IDS_REPEATED,
+                    Arrays.toString(repeatedIds.toArray())));
+        }
+        return ids;
+    }
+
+    private void validateActionArguments(final int obligationIndex, final List<ActionArgument> actionArguments) {
```
irinaepshteyn: Please add a unit test

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -102,14 +166,35 @@ private void validatePolicyActions(final Policy p) {
...
+                            action, this.validAcsPolicyHttpActions));
                 }
             }
         }
     }
 
+    private void vaLidatePolicyObligations(final List<String> obligationIds, final Policy policy) {
+        List<String> notFound = new ArrayList<>();
+        for (String obligationId : policy.getObligationIds()) {
+            if (StringUtils.isEmpty(obligationId)) {
```
irinaepshteyn: Should unload to schema validation if possible

service/src/test/java/com/ge/predix/acs/service/policy/admin/PolicyManagementServiceTest.java
irinaepshteyn: All these tests could be moved to PolicySetValidatorTest

irinaepshteyn (changes requested):
Done the first round.  Please review my comments.

model/src/main/java/com/ge/predix/acs/model/Obligation.java
sanjeevchopra: makes sense 👍

model/src/main/java/com/ge/predix/acs/model/Obligation.java
sanjeevchopra: I think we decided to not have that a field to indicate that. PEP can always chose to ignore it.

model/src/main/java/com/ge/predix/acs/model/PolicySet.java
sanjeevchopra: Re: type - we need one which does not allow duplicates and preserves order.

model/src/main/java/com/ge/predix/acs/model/Policy.java
sebastiantb: LinkedHashSet. implementing with linkedhashset

model/src/main/java/com/ge/predix/acs/model/PolicySet.java
sebastiantb: there is a validation for duplicated ids. I think order is only required on policy object when you set the obligationexpressionsids you want to use.

irinaepshteyn: For obligationExpressions contained in a policy set we should use Set to avoid duplicates and remove the need to explicitly validate for duplicates.  Order is not important.

For obligationIds contained per each policy we should use LinkedHashSet to avoid duplicates and preserve the order.  

Obligations returned in the policy evaluation result should be returned in the same order as obligationIds in the matching policy.  Thus for Obligations contained in evaluation result we should also use LinkedHashSet to avoid duplicates and preserve the order.

sebastiantb: ok!

irinaepshteyn: After further discussion I agree with @sebastiantb that it might be better to explicitly validate policy set for duplicate obligationExpressions upon upsert and return an error.  Deserializing a json with duplicate obligationExpressions into LinkedHashSet might produce non-intended obligation content.  

The same argument goes for obligationIds.  Deserializing into LinkedHashSet might produce non-intended order by silently replacing a duplicate ids.  For example, 1, 2, 3, 1 will be deserialized into 1,2,3.  What is intension was to keep 1 last?  It's better to explicitly validate for duplicates upon policy set upsert and return an error.

In that case using List is sufficient.  If we agree I will update the design to reflect that.

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
sebastiantb: yes, obligation id should be unique and with sets we cannot arbitrarily decide which one to keep as mentioned on the previous comment

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
sebastiantb: unit test are aready created, the will be moved to other class as specified

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
sebastiantb: private method it is being test on validatepolicyset call

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
sebastiantb: private method, being tested on validatepolicyset call

model/src/main/java/com/ge/predix/acs/model/ObligationExpressions.java
```diff
@@ -0,0 +1,120 @@
...
+@ApiModel(
+        description = "Obligation augments the interface contract between the policy engine (PDP) and "
+                + "enforcement point (PEP).  In addition to the actual access decision of either 'Permit' or 'Deny', "
+                + "the policy engine may return some supplementary information to the enforcement point in the form"
+                + " of an Obligation. Obligations are a set of operations that must be performed by the PEP "
+                + "in conjunction with an authorization decision, e.g., 'Permit, and in addition fulfill the "
+                + "indicated Obligation(s)' or, alternatively: 'Deny, and in addition fulfill the indicated"
+                + " Obligation(s)'.  If the PEP does not understand, or cannot fulfill, any of the obligations, "
+                + "then it MUST act as if the PDP had returned a 'Deny' authorization decision value.")
+public class ObligationExpressions {
```
irinaepshteyn: Should be singular - ObligationExpression.  Policy set will then include:
List<ObligationExpression> obligationExpressions;

model/src/main/java/com/ge/predix/acs/model/PolicySet.java
```diff
@@ -43,21 +49,35 @@ public PolicySet(final String name) {
...
+     */
+    public List<ObligationExpressions> getObligationExpressions() {
+        return this.obligationExpressions;
+    }
+
+    /**
+     * @param obligationExpressions
+     *            the obligationExpressions to set
+     */
+    public void setObligationExpressions(final List<ObligationExpressions> obligationExpressions) {
```
irinaepshteyn: Add unit tests to [PolicySetTest](https://github.com/predix/acs/blob/develop/model/src/test/java/com/ge/predix/acs/model/PolicySetTest.java) to validate correct deserialization including type being limited to CUSTOM only.

model/src/main/java/com/ge/predix/acs/model/ObligationExpressions.java
```diff
@@ -0,0 +1,120 @@
...
+
+    @NotNull(message = "id cannot be null or empty")
+    @Size(min = 1, message = "id cannot be null or empty")
+    private String id;
+
+    private ObligationType type = ObligationType.CUSTOM;
```
irinaepshteyn: Should be tested that this field is optional and set to CUSTOM if not provided in policy set upsert request body.  Should test if format exception is thrown if any value other than CUSTOM is provided.

model/src/main/java/com/ge/predix/acs/model/Policy.java
```diff
@@ -32,6 +33,8 @@
     private String name;
     private Target target;
     private List<Condition> conditions = Collections.emptyList();
+    private ArrayList<String> obligationIds = new ArrayList<String>();
```
irinaepshteyn: Why can't you use the same static Collections.emptyList() as for conditions?

model/src/main/java/com/ge/predix/acs/model/PolicySet.java
```diff
@@ -19,17 +19,23 @@
...
-@ApiModel(description = "A collection of access control policies evaluated in order. The first applicable policy"
-        + " determines the access control effect.")
+@ApiModel(
+        description = "A collection of access control policies evaluated in order. The first applicable policy"
+                + " determines the access control effect.")
 @SuppressWarnings("javadoc")
 public class PolicySet {
     private List<Policy> policies = new ArrayList<>();
+    @Valid
+    private List<ObligationExpressions> obligationExpressions = new ArrayList<>();
```
irinaepshteyn: Use static Collections.emptyList() to initialize

model/src/main/java/com/ge/predix/acs/model/ObligationExpressions.java
```diff
@@ -0,0 +1,120 @@
...
+                + "the policy engine may return some supplementary information to the enforcement point in the form"
+                + " of an Obligation. Obligations are a set of operations that must be performed by the PEP "
+                + "in conjunction with an authorization decision, e.g., 'Permit, and in addition fulfill the "
+                + "indicated Obligation(s)' or, alternatively: 'Deny, and in addition fulfill the indicated"
+                + " Obligation(s)'.  If the PEP does not understand, or cannot fulfill, any of the obligations, "
+                + "then it MUST act as if the PDP had returned a 'Deny' authorization decision value.")
+public class ObligationExpressions {
+
+    @NotNull(message = "id cannot be null or empty")
+    @Size(min = 1, message = "id cannot be null or empty")
```
irinaepshteyn: I didn't realize initially that these annotations are not jackson annotations and have nothing to do with JSON schema.  We already define a JSON schema for policy set - [acs-policy-set-schema](https://github.com/predix/acs/blob/develop/service/src/main/resources/acs-policy-set-schema.json).  It would be more consistent to add obligation constraints there.

However, as we discussed with @anubhavi25 and @sebastiantb, these annotations expose constraints to ACS model users.  They also allow us to perform obligations model validation after binding user input to it but before request enters the policy management controller (add @Valid annotation to the request body of createPolicySet call).  That removes the need to add obligation constraints to the policy set JSON schema and explicitly test it at the service level.  To be consistent with existing implementation, any constraint violations (ConstrainViolationException) should return  HttpStatus.UNPROCESSABLE_ENTITY.  For handling ConstrainViolationException refer to [global-error-handler-in-a-spring-rest-api](http://www.baeldung.com/global-error-handler-in-a-spring-rest-api)

The following constraints should be enforced on obligations:
- Policy:
-- ObligationIds - optional collection of strings. If provided, should contain minimum one id
- PolicySet: 
-- ObligationExpressions - optional collection of ObligationExpression.  If provided, should contain minimum one element
- ObligationExpression:
-- id - required, string, not null, non-empty
-- type - optional, string. If provided, should be CUSTOM
-- actionTemplate - required, string? object? not null, non-empty
-- actionArguments - optional collection of ActionArgument.  If provided, should contain minimum one element
- ActionArgument:
-- name - required, string, not null, non-empty
-- value - required, string, not null, non-empty

Obligation constraints should be unit tested.  For example refer to [unit-testing-validation-in-annotation-based-validating-spring-beans](http://lstierneyltd.com/blog/development/examples/spring-examples/unit-testing-validation-in-annotation-based-validating-spring-beans/)
Successful schema validation tests:
- policy set with at least one policy containing no obligationIds
- policy set containing no obligationExpressions
- policy set containing at least one obligationExpression with no type
- policy set containing at least one obligationExpression with no actionArguments
- policy set with a policy containing at least one obligationId and matching ObligationExpression with at least one actionArgument

Unsuccessful schema validation tests:
- policy containing at least one null/empty obligationIds
- policy set containing at least one null/empty obligationExpression
- policy set containing at least one obligationExpression with id absent/null/empty
- policy set containing at least one obligationExpression with type null/empty/not CUSTOM
- policy set containing at least one obligationExpression with absent actionTemplate
- policy set containing at least one obligationExpression with at least one null/empty actionArgument
- policy set containing at least one obligationExpression with at least one actionArgument with absent/null/empty name
- policy set containing at least one obligationExpression with at least one actionArguments with absent/null/empty value

At the service level we should only validate additional application level constraints: 
- repeated obligation ids in the policy
- duplicated obligation expressions in the policy set
- duplicated action arguments in the obligation expression
- orphan obligation ids in the policy (ids without matching obligation expression)

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -82,11 +91,78 @@ public void removeCachedConditions(final PolicySet policySet) {
 
     @Override
     public void validatePolicySet(final PolicySet policySet) {
+
+        ValidatorFactory factory = Validation.buildDefaultValidatorFactory();
+        Validator validator = factory.getValidator();
```
irinaepshteyn: To add  obligation constraints, either use javax annotations and @Valid in policy management controller (preferred) or use already defined JSON schema for policy set - [acs-policy-set-schema](https://github.com/predix/acs/blob/develop/service/src/main/resources/acs-policy-set-schema.json).  Then this additional validator is not needed.

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -82,11 +91,78 @@ public void removeCachedConditions(final PolicySet policySet) {
...
+        List<String> repeatedIds = new ArrayList<>();
+        int index = 0;
+
+        for (ObligationExpressions obligationExp : obligationExps) {  
+            
+            if (ids.contains(obligationExp.getId())) {
+                repeatedIds.add(obligationExp.getId());
+            }
+            //cannot validate with java x if it's an empty object, only null
+            if (((Map<Object, Object>) obligationExp.getActionTemplate()).isEmpty()) {
```
irinaepshteyn: Not null and non-empty action template could be enforced by setting "minLength":1 in policy set JSON schema.  Then this validation will be performed by validateSchema() above.  We should only explicitly validate here for duplicate obligation Ids.

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
sebastiantb: I don't think this is  possible, for list<string>obligationids, it can be empty or populated and the only validation is that there are not repeated or blank or empty string values, don't think javax validation annotations allows you to specify that the contents of a collection should not be null or blank, the annotations are applied to the collection itself

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -82,11 +91,78 @@ public void removeCachedConditions(final PolicySet policySet) {
...
+            sb.append("\n");
+        }
+        if (!violations.isEmpty()) {
+            throw new PolicySetValidationException(sb.toString());
+        }
+
+        validator.validate(policySet.getObligationExpressions());
+
         validateSchema(policySet);
+        List<String> obligationIds = validateObligationsAndGetIds(policySet.getObligationExpressions());
```
irinaepshteyn: Lets call it availableObligations

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -82,11 +91,78 @@ public void removeCachedConditions(final PolicySet policySet) {
...
+        List<String> duplicatedActionArguments = new ArrayList<>();
+        for (ActionArgument actionArgument : actionArguments) {
+            if (iteratedActionArguments.contains(actionArgument.getName())) {
+                duplicatedActionArguments.add(actionArgument.getName());
+            }
+            iteratedActionArguments.add(actionArgument.getName());
+        }
+        if (!CollectionUtils.isEmpty(duplicatedActionArguments)) {
+            throw new PolicySetValidationException(
+                    String.format(PolicySetValidationMessages.EXCEP_OBL_ACTION_ARG_DUPLICATED, obligationIndex,
```
irinaepshteyn: It would be better to reference obligation expression by id in the error message, not by index.

service/src/test/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorTest.java
```diff
@@ -229,4 +229,85 @@ public void testRemovalOfMultipleCachedConditions() {
...
+            expectedExceptionsMessageRegExp = ".*Obligation Expression validation failed.*id cannot be null or "
+                    + "empty at.*",
+            dataProvider = "obligationIdNullorBlank")
+    public void testValidatePolicyBadObligationIdNullorBlank(final String filename) {
+        PolicySet policySet = this.jsonUtils.deserializeFromFile(filename, PolicySet.class);
+        this.policySetValidator.validatePolicySet(policySet);
+    }
+
+    @DataProvider(name = "actionTemplateNullorBlank")
+    public Object[][] provideActionTemplateData() {
```
irinaepshteyn: Change to obligationActionTemplateNullorBlank and provideObligationActionTemplateData() so it's not mixed up with policy action.

service/src/test/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorTest.java
```diff
@@ -229,4 +229,85 @@ public void testRemovalOfMultipleCachedConditions() {
...
+            expectedExceptions = { PolicySetValidationException.class },
+            expectedExceptionsMessageRegExp = ".*Obligation Expression validation failed\\. actionTemplate"
+                    + " cannot be null or empty at.*")
+    public void testValidatePolicyBadObligationActionTemplateNullorBlank(final String filename) {
+        PolicySet policySet = this.jsonUtils.deserializeFromFile(filename, PolicySet.class);
+        this.policySetValidator.validatePolicySet(policySet);
+    }
+
+    @DataProvider(name = "actionArgumentNullorBlank")
+    public Object[][] provideActionArgumentData() {
```
irinaepshteyn: Change to obligationActionArgumentNullorBlank and provideObligationActionArgumentData()

service/src/test/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorTest.java
```diff
@@ -229,4 +229,85 @@ public void testRemovalOfMultipleCachedConditions() {
...
+    @Test(
+            expectedExceptions = { PolicySetValidationException.class },
+            expectedExceptionsMessageRegExp = ".*Obligation Expression validation failed\\. Obligation ids need to "
+                    + "be unique\\." + " Repeated values.*")
+    public void testValidateObligationIdsNotUnique() {
+        PolicySet policySet = this.jsonUtils
+                .deserializeFromFile("obligation/set-with-1-policy-bad-obligation-ids-repeated.json", PolicySet.class);
+        this.policySetValidator.validatePolicySet(policySet);
+    }
+
```
irinaepshteyn: Add the following successful schema validation tests:
- policy set with at least one policy containing no obligationIds
- policy set containing no obligationExpressions
- policy set containing at least one obligationExpression with no type
- policy set containing at least one obligationExpression with no actionArguments
- policy set with a policy containing at least one obligationId and matching ObligationExpression with at least one actionArgument

Add the following unsuccessful schema validation tests:
- policy set containing at least one null/empty obligationExpression
- policy set containing at least one obligationExpression with id absent/null/empty
- policy set containing at least one obligationExpression with type null/empty/not CUSTOM
- policy set containing at least one obligationExpression with absent actionTemplate
- policy set containing at least one obligationExpression with at least one null/empty actionArgument 
- policy set containing at least one obligationExpression with at least one actionArgument with absent name
- policy set containing at least one obligationExpression with at least one actionArguments with absent value

Additional policy set validation:
- policy set containing duplicated obligation expressions.

service/src/test/resources/obligation/set-with-1-policy-bad-action-template-blank.json
```diff
@@ -0,0 +1,87 @@
+{
```
irinaepshteyn: Policy sets for these tests don't have to be so elaborate.  If the use case in test is blank obligation action template then the policy set could be as simple as below:

{
  "name": "set-with-1-policy",
  "policies": [
    {
      "name": "Default",
      "effect": "PERMIT"
    }
  ],
  "obligationExpressions": [
    {
      "obligationId": "obligation1",
      "actionTemplate": {}
    },
    {
      "obligationId": "obligation2",
      "actionTemplate": {}
    }
  ]
}

Same applies to the rest of JSON files used for obligations unit tests.

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidationMessages.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private PolicySetValidationMessages() {
+    }
+
+    public static final String POLICY_VAL_FAILED = "Policy validation failed.";
+
+    public static final String OBL_VAL_FAILED = "Obligation Expression validation failed.";
+
+    public static final String EXCEP_OBL_IDS_REPEATED = OBL_VAL_FAILED + " Obligation ids need to be unique."
+            + " Repeated values:  %s";
```
irinaepshteyn: Change to " obligationExpressions ids have to be unique.  The following ids are repeated: [%s]"

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidationMessages.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public static final String EXCEP_OBL_ACTION_ARG_DUPLICATED = OBL_VAL_FAILED + " actionArguments names "
+            + "cannot be repeated on  index : [%d], values: [%s]";
```
irinaepshteyn: Error message should report obligation expression id (not index) where duplicate arguments are detected and duplicated arguments names.  " actionArguments names have to be unique.  Obligation expression [%s] contains the following repeated actionArguments: [%s]"

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidationMessages.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public static final String EXCEP_OBL_ACTION_TEMPLATE_NULL_EMPTY = OBL_VAL_FAILED + " actionTemplate"
+            + " cannot be null or empty at : [%s]";
```
irinaepshteyn: Change to " actionTemplate in obligation expression [%s] cannot be null or empty."

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidationMessages.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public static final String EXCEP_POLICY_OBL_IDS_NOT_FOUND = POLICY_VAL_FAILED + " The following "
+            + "obligationsIds do not match with the obligations defined for this policy set. [ policy: [%s]"
+            + ", obligationIds: %s ]";
```
irinaepshteyn: Change to " Unable to find matching obligation expression for the following obligationsIds defined in policy [%s]: [%s]."

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidationMessages.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public static final String EXCEP_POLICY_OBL_IDS_NULL_EMPTY = POLICY_VAL_FAILED + " obligationsIds cannot"
+            + " contain null or empty values for policy: [%s]";
```
irinaepshteyn: Change to " obligationIds in policy [%s] cannot contain null or empty values."

model/src/main/java/com/ge/predix/acs/model/PolicySet.java
```diff
@@ -19,17 +19,23 @@
...
  */
-@ApiModel(description = "A collection of access control policies evaluated in order. The first applicable policy"
-        + " determines the access control effect.")
+@ApiModel(
+        description = "A collection of access control policies evaluated in order. The first applicable policy"
+                + " determines the access control effect.")
 @SuppressWarnings("javadoc")
 public class PolicySet {
     private List<Policy> policies = new ArrayList<>();
+    @Valid
```
irinaepshteyn: What is @Valid for here?

irinaepshteyn (changes requested):
I finished reviewing model and policy set validation.  Please review my comments.

service/src/main/java/com/ge/predix/acs/service/policy/validation/PolicySetValidatorImpl.java
```diff
@@ -82,11 +91,78 @@ public void removeCachedConditions(final PolicySet policySet) {
 
     @Override
     public void validatePolicySet(final PolicySet policySet) {
+
+        ValidatorFactory factory = Validation.buildDefaultValidatorFactory();
```
irinaepshteyn: Default validator factory relies on some sort of bean validator implementation and expression language implementation present in the class path.  Currently, we happen to load hibernate bean validator version 5.3.5.Final (managed by cassandra-all dependency) and jasper apache-el version 8.0.33 (managed by spring-boot-starter-jetty 1.5.3.RELEASE).  

If we decide to go with JSR-303 bean validation annotations to enforce obligations schema and either use bean validator here or @Valid annotation at the controller layer, we have to explicitly specify implementation and version of bean validation and expression language and not rely on other dependencies.  Versions and implementations have to be consistent between ACS model  to use for constraints unit testing and ACS service to use for actual validation.

model/src/main/java/com/ge/predix/acs/model/ObligationExpressions.java
irinaepshteyn: Provided description is more suitable for Obligation class not ObligationExpression.  How about:

"Obligation Expressions are intended to allow a policy writer to add dynamic expressions in to the obligation statements.  When a policy is evaluated, any obligation expression included in that policy is evaluated to an obligation by resolving arguments of the action defined by the obligation expression.  Resolved obligations are then passed by the the policy engine (PDP) back to policy enforcement point (PEP) as part of policy evaluation result."

irinaepshteyn (approved):

irinaepshteyn:
@sebastiantb, it was decided to move this pull request from predix/acs to eclipse/keti.  Please hold off the review until it's done.  Thanks!

irinaepshteyn:
Pull request is moved to Eclipse Keti:  https://github.com/eclipse/keti/pull/4.  Closing...

PR #182, opened 2017-11-13
Removed predix specific functionality and tests

irinaepshteyn:
- Removed predix maven profiles and service startup scripts;
- Removed predix dependencies and cloud configurations;
- Removed ZAC monitoring, predix audit, metering, and dynamic log configurer;
- Removed dependency on cloud foundry utilities for testing;
- Removed predix specific integration tests;

PR #183, opened 2017-11-14
Refactored integration tests pom for easier merges with keti.

irinaepshteyn:


acs-integration-tests/pom.xml
```diff
@@ -96,22 +99,29 @@
                     </properties>
                     <systemPropertyVariables>
                         <ACS_SERVICE_ID>predix-acs</ACS_SERVICE_ID>
```
sebastiantb, line 101: does changing the service Id to to keti or acs break execution of tests?

sebastiantb (commented):
acceptance profile needs zac urls Does it need to run on predix profile or can it be run against a public profile? any chance we can have acceptance-public and acceptance-predix?

Looks good, can you add the pipeline jobs?

irinaepshteyn:
Test results:

https://predix1.jenkins.build.ge.com/job/Predix-Security/job/ACS/job/ACS-PullRequest/job/acs-integration-pullrequest/1089/
https://predix1.jenkins.build.ge.com/job/Predix-Security/job/ACS/job/ACS-PullRequest/job/acs-integration-pullrequest/1090/
https://predix1.jenkins.build.ge.com/job/Predix-Security/job/ACS/job/ACS-PullRequest/job/acs-integration-pullrequest/1092/
https://predix1.jenkins.build.ge.com/job/Predix-Security/job/ACS/job/ACS-PullRequest/job/acs-integration-pullrequest/1099/

acs-integration-tests/pom.xml
irinaepshteyn: Currently, acceptance profile is executed by the deploy pipeline and assumes predix and cloud foundry environment.  Not sure if we need acceptance tests for public version.  We could address it when the need arises.

irinaepshteyn: We could address service renaming and zone header (PREDIX_ZONE_ID) renaming at the time of renaming acs packages.

anubhavi25 (approved):
I would also review the comments in https://github.com/predix/acs/pull/182 to see if anything applies here but otherwise looks good 👍

irinaepshteyn:
@anubhavi25, introducing modules to integration tests so we could use nested poms (one per each profile), as suggested in PR#182, seems like an interesting idea and requires some investigation.  I will create a separate story for it and note that we have to be aware of common configurations and should duplicate them in each module as needed.

PR #187, opened 2017-12-08
added missing copyright

sanjeevchopra:


anubhavi25:
@sanjeevchopra thanks for adding this copyright header. I've also made changes to the script that will be used in the future to automatically update copyrights (most likely via Git hooks)

irinaepshteyn (approved):

